PR #4, opened 2015-08-17
Some small changes (warm up ;)

szborows:


freakboy3742:
Fantastic! Thanks so much for these contributions!

PR #7, opened 2015-08-23
Resolve Reference errors with r_long calls

glasnt:
Testing via testserver sample.py invocation
 - Before change: Uncaught ReferenceError: r_long is not defined
 - After change: no more uncaught error, but fails later in the sample.py

Tested a more limited test of changing sample.py to only includes `do_stuff` and `main`
 - Before change: Uncaught ReferenceError: r_long is not defined
 - After change: no error, expected output

### DIFF
--- a/batavia/modules/marshal.js
+++ b/batavia/modules/marshal.js
@@ -594,7 +594,7 @@ batavia.modules.marshal = {
             break;
 
         case batavia.modules.marshal.TYPE_TUPLE:
-            n = r_long(p);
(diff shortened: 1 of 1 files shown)


freakboy3742:
I think this bug was related to word size on the machine. My machine was marshalling tuples as "TYPE_SMALL_TUPLE"; yours was evidently using "TYPE_TUPLE". I'm guessing this is related to a 32-bit vs 64-bit Python install.

PR #9, opened 2015-09-15
Implement abs() builtin.

berkerpeksag:


freakboy3742:
Thanks for the contribution!

If you're looking for a "next step" contribution - I'd really like to see a full unit test suite for all the internals. Converting the "try builtins" into a "Run self-test suite" button that reports success and failure would be ideal.

berkerpeksag:
Thanks, I have a slightly different idea:

1. Write tests in separate Python files: We can even borrow some tests from CPython's test suite. Of course, simply copying and pasting them won't work, but we can at least take some test scenarios.
2. Auto generate `tests/test.html`: Currently, adding new tests is really hard. Creating the `<script id="batavia-TEST_NAME" type="application/python-bytecode">base64 encoded bytecode</script>` part automatically would be great. We can even use a tool like Gulp for this.

freakboy3742:
Ideally, we *would* be able to just drop in the PYCs from the Python test suite - however, I appreciate that this will be difficult, and is probably a long term goal.

As for your architectural approach - that sounds interesting. I'm not convinced gulp would be needed, though - surely you could generate the list dynamically using a template in index.html?

Ultimately, I'm really just aiming at having a test suite so that we can verify that the implementation is CPython compliant.

PR #14, opened 2016-03-10
Implement all, any, bool, hex, oct

glasnt:


batavia/VirtualMachine.js
```diff
@@ -156,7 +174,11 @@ batavia.builtins = {
     hasattr: function() { throw new batavia.builtins.NotImplementedError("Builtin Batavia function 'hasattr' not implemented"); },
     hash: function() { throw new batavia.builtins.NotImplementedError("Builtin Batavia function 'hash' not implemented"); },
     help: function() { throw new batavia.builtins.NotImplementedError("Builtin Batavia function 'help' not implemented"); },
-    hex: function() { throw new batavia.builtins.NotImplementedError("Builtin Batavia function 'hex' not implemented"); },
+    hex: function(args) { 
+        if (args.length !== 1)
+            throw new batavia.builtins.TypeError("hex() takes exactly one argument (" + args.length + " given)");
+	return args[0].toString(16);
```
lost-theory: do you need to prepend `0x`?

```
>>> hex(1)
'0x1'
```

vs.

```
> (1).toString(16)
'1'
```

Same goes for `oct` with a leading `0`.

glasnt: For oct, python 3.5 gives 
```
>>> oct(14)
'0o16'
```
where as python 2.7 gives 
```
>>> oct(14)
'016'
```

Which is preferred in this case, @freakboy3742?

freakboy3742: Py3 behavior is preferred. The rest of the syntax is Py3.

glasnt: Ack. The PR has been updated.

PR #18, opened 2016-04-02
Safely use temporary files

MCGallaspy:
I encountered some issues running the test server! This should clear things up. See the commit message:

```
Some operations on open files in Unix don't raise errors,
but they do on Windows! In this case, we must explicitly
handle tempfile opening/closing to avoid trouble, since
the otherwise nicer NamedTemporaryFile and friends can't guarantee
access again after a file is closed.

For example see the docs for os.remove.
```

Here's the relevant part from the os.remove docs, which seems to be true of moving files as well:

> On Windows, attempting to remove a file that is in use causes an exception to be raised; on Unix, the directory entry is removed but the storage allocated to the file is not made available until the original file is no longer in use.

### DIFF
--- a/testserver/urls.py
+++ b/testserver/urls.py
@@ -8,14 +8,17 @@ from django.shortcuts import render
 
 
 def bytecode(sourcefile):
-    with tempfile.NamedTemporaryFile() as temp:
-        py_compile.compile(sourcefile, cfile=temp.name)
-        with open(os.path.join(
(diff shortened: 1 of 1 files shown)


freakboy3742:
Where were you 6 months ago - I'm pretty sure this is the problem that was causing me all sorts of headaches trying to deploy a Batavia demo on Azure :-) Thanks for the contribution!

PR #21, opened 2016-04-02
implemented basic enumerate function

MisterRios:


batavia/core/builtins.js
```diff
@@ -148,8 +148,13 @@ batavia.builtins.divmod = function(args) {
     return [div, rem];
 };
 
-batavia.builtins.enumerate = function() {
-    throw new batavia.builtins.NotImplementedError("Builtin Batavia function 'enumerate' not implemented");
+batavia.builtins.enumerate = function(args) {
+    var result = [];
```
freakboy3742: Given that this should be a generator, not a list, it would be good to flag this in a comment - "FIXME: this should return a generator."

PR #25, opened 2016-04-02
Add hasattr -- incomplete, feedback wanted

phildini:
This code change makes `hasattr` work for the happy path, but I can't get it to print any errors to the console or the "stdout" when the arguments are wrong. If I could get some feedback on what I might be doing wrong, much obliged. Once the happy and failure paths are working, I'll add tests to test.html and sample.py

### DIFF
--- a/batavia/core/builtins.js
+++ b/batavia/core/builtins.js
@@ -182,10 +182,11 @@ batavia.builtins.frozenset = function() {
 };
 
 batavia.builtins.getattr = function(args) {
-    try {
-        return args[0][args[1]];
-    } catch (err) {
-        if (args) {
+    if (args) {
+        var attr = args[0][args[1]];
+        if (attr !== undefined) {
+            return attr
+        } else {
             if (args.length === 3) {
                 return args[2];
             } else if (args.length === 2) {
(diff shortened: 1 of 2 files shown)


phildini:
Using `getattr` to follow what's listed here: https://docs.python.org/3.5/library/functions.html#hasattr

glasnt:
What was the code you were running when you were having issues invoking errors? If you add those to the `testserver/sample.py` in the `try_builtins` part, I can have a look (and then once these run true, you have your tests :D)

phildini:
@glasnt I've updated the commit with some tests in sample.py -- when I run sample.py in batavia, lines 55 and 56 don't print anything after the command.

glasnt:
Looks like is an issue with `getattr`: addressing an invalid key in an object doesn't raise an error in JavaScript, and so the catch block in getattr isn't being called, so your hasattr catch block won't be called. 

 My guess is changing the sadpath to invoke on returning `undefined` as opposed to a catch block will fix this, in both `getattr` and `hasattr`

phildini:
Updated the commit to check for undefined. I couldn't leave the "known failure" in there, because it would stop execution.

glasnt:
This looks :shipit: to me; however, there is a problem with the pystone test. Unsure if this is you or your now older fork of master. 

If @freakboy3742 says ok, I'm happy to merge this and fix pystone off master

glasnt:
And it turns out that hidden under the fold of the PR merge button is doco for me to test your patch against master, which totally works for pystone, so :+1:

PR #28, opened 2016-04-02
Build a table mapping opcodes to method calls

gasman:
This avoids the need to do string munging on opcode names for each instruction,
leading to a ~25% speed improvement in PyStone.


freakboy3742:
MOAR SPEED! This is incredible! Thanks!

PR #34, opened 2016-04-03
Added missing package.json

MarkusH:


freakboy3742:
The test framework has changed a bit since this was proposed; closing on the basis that it's no longer relevant.

PR #37, opened 2016-04-03
RFC: Cleanup Builtins, with default template doco

glasnt:
I propose a psudo-standard format for the javascript builtins:

As per the 'general format' doco at the top of the file: 
 * prepend the Python 3.4 `help(<fn>)` doco, for reference
 * test for expected array lengths
 * tests for types, if that's a mandatory
 * then implement the things

The general idea is to ensure that we're capturing the same functionality as the base python, where possible (types are FUN)

Also, this PR removes a few functions that don't exist in Python 3.4. The original list may have been bulkimported from Python 2.7 (at a guess), so my theory is that if `<fn>()` `NameError`s, and `help(<fn>)` also errors, then it should be removed. 

Now, *how* this is all done is contentious. Where possible, should raw JS be used in all the builtins, as not to create nested errors? The issues with `hasattr` and `getattr` show that there are some cases where using nested calls are useful, but I'm personally of the opinion that using JS length as opposed to Batavia length is the go. 

Comments welcome

### DIFF
--- a/batavia/core/Exception.js
+++ b/batavia/core/Exception.js
@@ -12,6 +12,9 @@ batavia.builtins.BaseException.prototype.toString = function() {
     }
 };
 
+batavia.builtins.ArgumentsError = function(msg) {
+    batavia.builtins.BaseException.call(this, 'ArgumentsError', msg);
+};
 batavia.builtins.ArithmeticError = function(msg) {
     batavia.builtins.BaseException.call(this, 'ArithmeticError', msg);
 };
--- a/batavia/core/builtins.js
+++ b/batavia/core/builtins.js
@@ -40,11 +40,17 @@ batavia.builtins.__import__ = function(args, kwargs) {
 
 General builtin format: 
 
-batavia.builtins.<fn> = function(<args>) { 
-    // length checks - is it expected that there's only a specific number of arguments?
-    if (args.length !== <length>) {
-        throw new batavia.builtins.TypeError(
-            "<fn>() takes exactly <length> argument(s) (" + args.length + " given)");
+
+// Example: a function that accepts exactly one argument, and no keyword arguments
+batavia.builtins.<fn> = function(<args>, <kwargs>) { 
+    if (arguments.length != 2) {
+        throw new batavia.builtins.RuntimeError("Batavia calling convention not used.")
+    }
+    if (kwargs && Object.keys(kwargs).length > 0) {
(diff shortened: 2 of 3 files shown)


glasnt:
New update: instead of help(<fn>), use <fn>.__doc__ as code-as-documentation. 

Untested, because currently sample.py in master is returning other errors that prevent local testing of things.

freakboy3742:
This all looks good to me - no disagreements on the general approach being proposed. 

Two small suggestions:
* we may want to enforce a general prototype of `(args, kwargs)`, even when the method doesn't take any kwargs
* we may want to check `arguments.length` to make sure that it's always 2, just to avoid any `map(parseInt)`-style argument mapping bugs. There shouldn't be any way a Python method can be invoked with anything but 2 arguments, but just in case...

As for raw JS vs Batavia methods - I think the dividing line is to know what type of object you're working with. If it's a JS object, use the native JS method; if it's a Batavia object, use the Python API (which might be deferring to a raw JS object, but might be overridden). In the case of `args`/`charges` - they're both constructed as part of the function dispatch process, so they're guaranteed to be JS objects, so raw JS methods are appropriate. The data *contained* in `args`/`kwargs`, however, will generally be Batavia objects, so Batavia methods should be used.

glasnt:
Do you mean a general structure of: 
```
batavia.builtins.abs = function(params) {
  if (params.length !== 2) { throw new batavia.builtins.ArgumentsError} 
  [args, kwargs] = params
```

note: using params instead of args to denote the args/kwargs vs params
The third line is possibly ['new' JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) but can be replaced with just a more verbose version

freakboy3742:
That would do the job, but I was thinking something more like:

    batavia.builtins.abs = function(args, kwargs) {
        if (arguments.length != 2) {
            throw new batavia.core.RuntimeError('Batavia calling convention not used.')
        }
        if (kwargs || kwargs.length > 0) {
            throw new batavia.core.ArgumentsError('abs doesn't accept keyword arguments')
        }
        if (!args || args.length != 1) {
            throw new batavia.core.ArgumentsError('abs expected exactly 1 argument (got X)')
        }

glasnt:
Done, with some minor alterations on your suggested code.

with this change, sample.py runs through to `abs(None)` (known failure line)

PR #43, opened 2016-04-03
Small optimisations

funkybob:
- Hoist batavia.modules.dis into function namespace in batavia.VirtualMachine.prototype.parse_byte_and_args to avoid lookups

- micro-optimise two loops


glasnt:
:shipit:

PR #49, opened 2016-04-18
Dict implementation, restoring js helper utils

andrzejkrawczyk:
- restored js helper utils for builting functions, isType, fixing broken builtin function sorted
- added basic Tuple implementation for **kwargs and list of tuples

### DIFF
--- a/batavia/core/builtins.js
+++ b/batavia/core/builtins.js
@@ -348,8 +348,42 @@ batavia.builtins.delattr = function(args, kwargs) {
 batavia.builtins.delattr.__doc__ = "delattr(object, name)\n\nDelete a named attribute on an object; delattr(x, 'y') is equivalent to\n``del x.y''.";
 
 batavia.builtins.dict = function(args, kwargs) {
-    throw new batavia.builtins.NotImplementedError(
-        "Builtin Batavia function 'dict' not implemented");
+    if (args.length > 1) {
+        var errorMsg = "dict expected at most 1 arguments, got " + args.length;
+        throw new batavia.builtins.TypeError(errorMsg);
+    }
+
+    var result;
+    if (args[0] !== undefined) {
+        result = {};
+        args[0].forEach(function (element, index) {
+            if (!element instanceof batavia.core.Tuple) {
+                throw new batavia.builtins.NotImplementedError("Builtin Batavia function 'dict' implemented only for list of Tuples");
+            }
+
+            if (element.length !== 2) {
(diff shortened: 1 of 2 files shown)


batavia/utils.js
```diff
@@ -684,3 +684,28 @@ batavia.make_callable = function(func) {
...
+
+/*************************************************************************
+ * utils
+ *************************************************************************/
+
+batavia.utils = batavia.utils || {};
+batavia.utils.isType = function isType(type, obj) {
+    var clas = Object.prototype.toString.call(obj).slice(8, -1);
+    return obj !== undefined && obj !== null && clas === type;
+};
```
freakboy3742: I took these out for two reasons - firstly, they're not needed, and secondly, they're based on a string comparison, so they're not reliable anyway.

As far as I can make out, you can do reliable type checks for everything we're likely to need with `instanceof` - if there's a case where this doesn't work, let me know and I can dig into this a bit deeper.

batavia/core/builtins.js
```diff
@@ -348,8 +348,42 @@ batavia.builtins.delattr = function(args, kwargs) {
 batavia.builtins.delattr.__doc__ = "delattr(object, name)\n\nDelete a named attribute on an object; delattr(x, 'y') is equivalent to\n``del x.y''.";
 
 batavia.builtins.dict = function(args, kwargs) {
-    throw new batavia.builtins.NotImplementedError(
-        "Builtin Batavia function 'dict' not implemented");
+    if (args.length > 1) {
```
freakboy3742: The top of builtins.js contains a template for builtin functions to follow; there's a couple of edge cases here that you haven't covered.

andrzejkrawczyk: isType is used in builtin sorted, taking that out will break the function.

Also instanceof is good enough only for checking custom types with custom constructors.
If you need to check JavaScript type this is totally broken. 
More about this with examples in this article:
http://bonsaiden.github.io/JavaScript-Garden/#types.equality

Should i Still remove isType util and rework sorted function in builtins?

batavia/utils.js
andrzejkrawczyk: isType is used in builtin sorted, taking that out will break the function.

Also instanceof is good enough only for checking custom types with custom constructors.
If you need to check JavaScript type this is totally broken. 
More about this with examples in this article:
http://bonsaiden.github.io/JavaScript-Garden/#types.equality

Should i Still remove isType util and rework sorted function in builtins?

freakboy3742:
Closing this PR - #50 implemented dict().

PR #50, opened 2016-04-18
Implement dict builtin

MisterRios:


MisterRios:
I've been working on dict for a while and recently saw that error handling is first and separated out, so I rewrote it, and it's much cleaner.

However, I'm having trouble on two things:
1) Passing a dict into dict() should return the original dict. It's returning it, but calling up an indexed value is not returning the correct value.
    
    predict = dict({"red": 2})
    print(predict)  # prints {red:2}
    print(predict["red"] == 2)  # prints false
    

2) When running the tests, this one fails in the CL test suite, and in the batavia testbed app:
    `dict((1, 2.345, "other"))` throws `TypeError: p.refs is undefined` in the batavia testbed app.
    in the CL testsuite throws 
```
test_tuple (tests.builtins.test_dict.BuiltinDictFunctionTests) ... FAIL

FAIL: test_tuple (tests.builtins.test_dict.BuiltinDictFunctionTests)
  File "/home/misterrios/Projects/batavia/tests/utils.py", line 729, in func
    self.assertBuiltinFunction(x=example, f=function, operation=operation, format=self.format)
  File "/home/misterrios/Projects/batavia/tests/utils.py", line 749, in assertBuiltinFunction
    """ % kwargs, "Error running %(operation)s with x=%(x)s" % kwargs)
  File "/home/misterrios/Projects/batavia/tests/utils.py", line 448, in assertCodeExecution
AssertionError: "TypeError: undefined is not an object (e[380 chars]:1\n" != '### EXCEPTION ###\nTypeError: cannot con[69 chars]:3\n'
+ ### EXCEPTION ###
+ TypeError: cannot convert dictionary update sequence element #0 to a sequence
+     test.py:3
- TypeError: undefined is not an object (evaluating 'p.refs.push')
-   phantomjs://code/batavia.min.js:35 in r_ref
-   phantomjs://code/batavia.min.js:45 in r_object
-   phantomjs://code/batavia.min.js:56 in r_object
-   phantomjs://code/batavia.min.js:77 in r_object
-   phantomjs://code/batavia.min.js:83 in read_object
-   phantomjs://code/batavia.min.js:85 in load_pyc
-   phantomjs://code/batavia.min.js:263 in run
-   :1
-   :1
 : Global context: Error running f(x) with x=(1, 2.3456, 'another')

Ran 7 tests in 1.430s
```

freakboy3742:
1. Sounds like something is getting an incomplete copy somewhere along the way. There are two places I'd suggest to set a breakpoint - 
  
  * in [utils.js, around line 467](https://github.com/pybee/batavia/blob/master/batavia/utils.js#L467) - this is the implementation of the subscript operator, so you can see if the object `predict` has the right keys at the point that you are requesting `red`
  * in [VirtualMachine.js, at line 548](https://github.com/pybee/batavia/blob/master/batavia/VirtualMachine.js#L548) - this is the "run the next opcode" command, so you can step through the Python code instruction by instruction to see how the object is being created.
  * in [VirtualMachine.js, at line 647](https://github.com/pybee/batavia/blob/master/batavia/VirtualMachine.js#L743) - this is the implementation of the `LOAD_NAME` opcode, which pulls data out of the locals dictionary. This will allow you to check what object is being retrieved.
  * in [VirtualMachine.js, at line 662](https://github.com/pybee/batavia/blob/master/batavia/VirtualMachine.js#L743) - this is the implementation of the `STORE_NAME` opcode, which puts data into the locals dictionary. This will allow you to check what object is being stored.

2. That error usually indicates that something is going wrong in the parsing process. Either the test code has an error in it, or there's a data type that isn't being parsed correctly. I'll take a look at this, but if you want to poke around yourself, A breakpoint in [marshal.js, at line 311](https://github.com/pybee/batavia/blob/master/batavia/modules/marshal.js#L311) should give you a pretty good idea what is going on; the problem test case will be one of `dict(1)`, `dict(2.345)`, or `dict('another')`.

MisterRios:
1. Haha! I was returning the array containing the dictionary instead of the dictionary itself. This has been corrected.

2. I did add the error for passing a float because of your comments. I'd forgotten about that. However, it looks like the error is coming from mixed types in tuples. I get the same error from `print((5.2, 5))` for example. Any chance you could take a look?

2a. Is there any type checker that has been built in? It would make the error handling much easier. In builtins.js line 846 there is a reference to a typechecker: `if (batavia.utils.isType(batavia.utils.TYPES.ARRAY, iterable))` but I can't find the reference in utils.js and it doesn't seem to work anyhow.

Thanks again.

freakboy3742:
I can confirm I'm getting the same error on `print((5.2, 5))`; I'm taking a look now.

The `isType` call is an artefact of some old code. You should be able to replace that with `iterable.prototype === Array`. I'll take a look at this too (if you don't beat me to it :-).

freakboy3742:
I've got to the bottom of this - there was some leftover code from an older fix for marshalling floats that was getting activated, causing problems. I've applied the fix in 2ed63744 - which has, in the process, cleaned up a couple of tuple-related tests that were failing. Thanks a bunch for narrowing down the problem to `(5.2, 5)` - that was incredibly helpful.

I've also corrected the usage of `isType` in 5f64c434. The implementation of `sorted()` still needs a bit of work, but "working for a limited number of cases" is better than "not working at all".

From what I can tell, once you've merged these two commits into your branch, your implementation of `dict()` should be pretty much ready to go.

MisterRios:
Normally one would celebrate an 'unexpected success' (let alone three!). I'll look at where the problem lies...

freakboy3742:
Looks like a pretty easy fix - the test suite is reporting "3 unexpected successes", which means *your* code is doing the right thing, but the test suite if flagging 3 tests as expected failures. The three tests that are passing unexpectedly all make sense as new passes - they all relate to dict construction, and 2 of them involve tuples.

So - if you delete the "test_int" line from the `not_implemented` list of:
* tests.builtins.test_dict.BuiltinDictFunctionTests

And remove the `@expectedFailure` decorator from:
* tests.datatypes.test_dict.DictTests.test_builtin_constructor
* tests.datatypes.test_dict.DictTests.test_builtin_non_2_tuples

It looks like your patch will be good to go!

freakboy3742:
Fantastic! Thanks for the patch!

PR #52, opened 2016-04-25
ubuntu 16.04 makes things better

eichin:
I've only checked versions, not tested (yet) but Ubuntu 16.04 shipped last week (including for Raspi: https://wiki.ubuntu.com/ARM/RaspberryPi#Ubuntu_16.04_LTS ) and while upgrading *just* for pybee might be a bit of a stretch, this document will last well into the 16.04 lifecycle...


freakboy3742:
Thanks for the updated info - nice to see that our lives are getting easier with time :-)

Issue #53, opened 2016-05-05
str() takes at most 1 argument (2 given)


FeralBytes:
Essentially I just tried to do the tutorial and putting 
`print('Hello World')`
into the text area and clicking run immediately resulted in a trace-back. Is this due to django running as python 2. Because the Shebang says python not python3.

By the way this is an awesome concept and so small and light weight! I am going to start testing if I can do AJAX using it; has this already been tested; are there some examples?

FeralBytes:
[I made a pull request here](https://github.com/FeralBytes/batavia/pull/1)
Please review it and let me know which method you would like to use to detect python3 and I will make a proper pull request against your project.

FeralBytes:
OK new pull request created using the preferred method; sorry for the multiple commits; I can pull it back and make a single commit if you would prefer? I did not test it until after the first commit.

freakboy3742:
Github lets me collapse multiple-commit pull requests into a single commit with a single button press, so don't worry about collapsing. There are a couple of comments against the patch; if you can address those, we can merge!

FeralBytes:
Done.

PR #54, opened 2016-05-06
Try to ensure manage.py is launched using py3.

FeralBytes:
Will resolve issue #53 .

### DIFF
--- a/testserver/manage.py
+++ b/testserver/manage.py
@@ -3,6 +3,14 @@ import os
 import sys
 
 if __name__ == "__main__":
+    
+    if 'python3' not in sys.executable:  # might be better to check version also.
+    # if sys.hexversion < 0x03000000:
+        python3_path = distutils.spawn.find_executable('python3')
+        if python3_path is not None:
+            os.execvp(sys.executable + '3', [sys.executable + '3'] + sys.argv)
+            #TODO what to do if it is None?
+    
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "settings")
 
     from django.core.management import execute_from_command_line
--- a/testserver/manage.py
+++ b/testserver/manage.py
@@ -6,7 +6,7 @@ if __name__ == "__main__":
     
     if 'python3' not in sys.executable:  # might be better to check version also.
     # if sys.hexversion < 0x03000000:
-        import distutils
+        import distutils.spawn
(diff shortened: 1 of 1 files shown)


testserver/manage.py
```diff
@@ -3,6 +3,15 @@
 import sys
 
 if __name__ == "__main__":
+    
+    if 'python3' not in sys.executable:  # might be better to check version also.
```
funkybob: Why not:

    if sys.version_info[0] != 3:

testserver/manage.py
```diff
@@ -3,6 +3,15 @@
...
+    # if sys.hexversion < 0x03000000:
+        import distutils.spawn
+        python3_path = distutils.spawn.find_executable('python3')
+        if python3_path is not None:
+            os.execvp(sys.executable + '3', [sys.executable + '3'] + sys.argv)
+            #TODO what to do if it is None?
```
freakboy3742: What to do:
```
print('Batavia requires Python 3', file=sys.stderr)
sys.exit(1)
```

FeralBytes: `print('Batavia requires Python 3', file=sys.stderr)`
Is not compatible with Python 2. So I changed it too:
`sys.stderr.write('Batavia requires Python 3' + os.linesep)`

testserver/manage.py
FeralBytes: Good point; I had not arrived at that solution; but I like it better. Implemented.

testserver/manage.py
```diff
@@ -3,6 +3,16 @@
...
 
 if __name__ == "__main__":
+    
+    if sys.version_info[0] != 3:
+        import distutils.spawn
+        python3_path = distutils.spawn.find_executable('python3')
+        if python3_path is not None:
+            os.execvp(sys.executable + '3', [sys.executable + '3'] + sys.argv)
+        else:
+            sys.stderr.write('Batavia requires Python 3' + os.linesep)
```
freakboy3742: The spawn check isn't needed - having python2 start python3 is just going to lead to confusion. If sys.version_info doesn't return the right result, it's better to throw an error and let the user deal with that.

FeralBytes: OK. I disagree; but it is your project.

freakboy3742:
Thanks for the contribution! Hopefully nobody else will get bitten by accidentally using Python2.

Issue #47, opened 2016-04-13
Complete implementations of Python builtins


freakboy3742:
`batavia/core/builtins.js` is an implementation of all the Python builtins - methods like `print()`, `input()`, `list()`, `abs()`, `min()` and so on.

An implementation of these builtins is required.

schuyler1d:
i'm going to implement filter()

rightlag:
I'm going to work on `reversed()` 💥

lisroach:
I will implement `list()`

schuyler1d:
also doing map()

thewus:
I'm going to work on len()

Brigette-Eckert:
Going to work on this issue,  is there any particular method that is needed ASAP? otherwise I'll pick one and post.

sethMittag:
working on set()

LawlietBlack:
working on oct()

Brigette-Eckert:
working on next()

LawlietBlack:
working on all()

Brigette-Eckert:
working on isinstance()

prathmesh4:
I m work beginning

aeroaks:
Working on input().

asherScott:
Which builtins have yet to be implemented?

freakboy3742:
@asherScott At the very least, the methods in [this file](https://github.com/pybee/batavia/blob/master/batavia/core/builtins.js this file) whose implementation is nothing more than `throw NotImplmentedError`. There are a couple of methods that *internally* throw `NotImplementedError` as well, for certain subsets of behavior (e.g., `iter` with 2 arguments).

greeve:
I am working on implementing ascii in batavia/core/builtins.js.

candeira:
I'm dibbing byte, bytearray, and adding to Byte and Bytearray, which depend on each other enough that it doesn't make sense to work on one and not the others. 

I've opened a (so far) empty PR to track progress of my work: https://github.com/pybee/batavia/pull/221

candeira:
Progressing on `bytes`  by vendoring the Buffer JS library, stalled on `bytearray` because Buffer is mutable, but fixed-length. I'll keep reporting.

shyam114:
Hey @freakboy3742 the link in this (https://github.com/pybee/batavia/issues/47#issuecomment-230678203) points to 404 error page.

freakboy3742:
@shyam114 Thanks - I've just corrected the link.

[...]

PR #55, opened 2016-05-07
Implement builtin float method

pietertolsma:
Implemented the float method. Part of ticket #47 . Also add `env` to .gitignore.

Disclaimer: this is my first time contributing to an open-source project.

### DIFF
--- a/batavia/core/builtins.js
+++ b/batavia/core/builtins.js
@@ -475,29 +475,29 @@ batavia.builtins.filter = function() {
 };
 
 batavia.builtins.float = function(args) {
-    if(args.length > 1){
+    if (args.length > 1) {
         throw new batavia.builtins.TypeError("float() takes at most 1 argument (" + args.length + " given)");
     }
-    if(args.length === 0){
+    if (args.length === 0) {
         return 0.0;
     }
 
     var toConvert = args[0];
 
-    if(typeof(toConvert) === "string"){
-        if(toConvert.search(/[^0-9.]/g) === -1){
+    if (typeof(toConvert) === "string") {
+        if (toConvert.search(/[^0-9.]/g) === -1) {
             return parseFloat(toConvert);
-        }else{
-            if(toConvert === "nan" || toConvert === "+nan" || toConvert === "-nan"){
+        } else {
+            if (toConvert === "nan" || toConvert === "+nan" || toConvert === "-nan") {
                 return NaN;
-            }else if(toConvert === "inf" || toConvert === "+inf"){
+            } else if(toConvert === "inf" || toConvert === "+inf") {
                 return Infinity;
-            }else if(toConvert === "-inf"){
+            } else if(toConvert === "-inf") {
                 return -Infinity;
             }
             throw new batavia.builtins.ValueError("could not convert string to float: '" + args[0] + "'");
         }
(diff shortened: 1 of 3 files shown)


pietertolsma:
Somehow it is not passing the tests. `Can't find variable; batavia`. I don't see how my changes affected this?

jdorfman:
@freakboy3742 any chance you can check this out for @pietertolsma and possibly point him the right direction? Thanks!

freakboy3742:
@pietertolsma Thanks for the patch! The failure is an odd one - it looks like it's a system-wide failure caused by the minified Batavia script not being compiled. 

Can you confirm that you've been able to run the test suite locally, and if so - on what platform? Can you also provide the output of `pip freeze`, and the version of PhantomJS that you're using in your local testing.

pietertolsma:
@freakboy3742 Weird, my local test suite is giving the same results as Travis CI (I'm running OS X El Capitan).

`pip freeze` gives me:
```-e git+https://github.com/pietertolsma/batavia.git@ffa87affce7eee25cff416d55c1399cadfa8b72b#egg=batavia
jsmin==2.2.1```

I'm using version 2.1.1 of PhantomJS.

**Comment by pietertolsma** [comment]
Wow. Just noticed what's wrong. Syntax error (typo) in the method. Uploading fix now.

File: batavia/core/builtins.js [hunk 1]
```diff
@@ -474,10 +474,36 @@ batavia.builtins.filter = function() {
...
+    if(args.length > 1){
+        throw new batavia.builtins.TypeError("float() takes at most 1 argument (" + args.length + " given)");
+    }
+    if(args.length === 0){
+        return 0.0;
+    }
+
+    var toConvert = args[0];
+
+    if(typeof(toConvert)) === "string"){
```
pietertolsma: Typo here

pietertolsma:
So no failures this time, but an unexpected success. Do I need to edit a test case?

freakboy3742:
@pietertolsma Yes - the "unexpected success" indicates a test that now passes, whereas previously it didn't. It's entirely possible (and probable) that by fixing one specific operator, a couple of other tests now pass as well. For example, if you implement the "+" operator, "+=" will probably also start working, because the implementation of "+=" falls back on "+" by default. 

When you run the test suite, it will tell you the name of the test (or tests) that are now passing. Find the test file (tests/datatypes/...) that corresponds to the unexpected success. In that file, you'll find test cases, each of which have a "not implemented" list. This details the tests that are known to pass because the underlying operator isn't implemented. Remove the lines from that list that correspond to the unexpected success, save, and re-run the test suite. You should now get a test suite that passes with no failures, and no unexpected successes. Commit and push, and the CI server should re-run the test suite for you; once that is confirmed, I can merge the patch!

One other request if you get a chance: we prefer to have spaces around if clauses - i.e.
```
if (x > y) {
   ...
} else if (x < z) {
  ...
} else {
   ...
}
```
It's a little thing, but keeping the code style consistent aids readability in the long term.

pietertolsma:
Cheers! @freakboy3742 thank you for all the help! Hope to contribute more to this project as time passes 👍

freakboy3742:
Fantastic! Thanks Pieter - I look forward to seeing your next contribution!

PR #56, opened 2016-05-09
test_add_list

betojulio:
I change the error message to pass the test case.


betojulio:
Hi Russ, I just wanted to thank you for the help provided. It is amazing
to  contribute to batavia. I will try to be more independent from now on.

I just have question, the implementation will work for 'string' + [] but
will not work for [] + 'string', is this ok or you will work in a test file
that check this.

Thanks.

freakboy3742:
Hi Julio,

You’re most welcome - thanks for contributing!

It’s completely OK if a something works one way, but not the other - as
long as your fix doesn’t break anything that already works. The test suite
will validate both directions: test_str.test_add_list is a different test
to test_list.test_add_string. After your patch, the former now passes, but
the latter is still throwing an error. If you submit a patch to fix the []
+ “string” case, you just have to make sure you don’t break any other tests
along the way.

Yours,
Russ Magee %-)

PR #58, opened 2016-05-10
test_add_none implemented.

betojulio:
test_add_none implemented.


betojulio:
I believe it does not pass because It is 2 different commits I will close the pull and make only one commit.

freakboy3742:
The fact that it's two different commits doesn't matter - the test suite will always run on the "head" of any branch. What has happened is that by fixing Str.test_add_none, you've inadvertently fixed two other tests as well:
```
test_multiply_none (tests.datatypes.test_bytes.BinaryBytesOperationTests) ... unexpected success
test_multiply_none (tests.datatypes.test_bytes.InplaceBytesOperationTests) ... unexpected success
```

If you delete the lines related to to these two tests and update the PR, the test suite should give you a green tick again.

betojulio:
mmm, ok, but that is the second commit that I did 'test_multiply_none' .

betojulio:
OOOhh now I see what you meant I was working on test_str not on test_bytes .  I will update PR # 59

PR #59, opened 2016-05-11
test_add_none for string ,test_multiply_none for string and test_multiply_none for bytes are implemented.

betojulio:


betojulio:
finally!!!

PR #63, opened 2016-05-12
Implement builtin input method

pietertolsma:
Input is asked in console. @freakboy3742 is this the right way of doing this? Thanks in advance!


batavia/core/builtins.js
```diff
@@ -575,10 +574,18 @@ batavia.builtins.id = function() {
...
+    if (args.length > 1) {
+      throw new batavia.builtins.TypeError("input expected at most 1 arguments, got " + args.length);
+    }
+    if (args.length === 1) {
+      batavia.stdout(String(args[0]) + "\t");
+    }
+    return readline();
 };
 
+batavia.builtins.input.__doc__ = "input([prompt]) -> Retrieves input from user in console. Prompt argument is optional. ";
```
freakboy3742: This string should match the one returned by Python itself as `input.__doc__` - in this case: 
`'input([prompt]) -> string\n\nRead a string from standard input.  The trailing newline is stripped.\nIf the user hits EOF (Unix: Ctl-D, Windows: Ctl-Z+Return), raise EOFError.\nOn Unix, GNU readline is used if enabled.  The prompt string, if given,\nis printed without a trailing newline before reading.'`

batavia/core/builtins.js
```diff
@@ -575,10 +574,18 @@ batavia.builtins.id = function() {
     throw new batavia.builtins.NotImplementedError("Builtin Batavia function 'id' not implemented");
 };
 
-batavia.builtins.input = function() {
-    throw new batavia.builtins.NotImplementedError("Builtin Batavia function 'input' not implemented");
+batavia.builtins.input = function(args) {
+    if (args.length > 1) {
+      throw new batavia.builtins.TypeError("input expected at most 1 arguments, got " + args.length);
+    }
+    if (args.length === 1) {
```
freakboy3742: There should also be a kwargs check here to make sure no kwargs have been provided.

freakboy3742:
@pietertolsma `readline()` would make sense for Node, but something like `window.prompt("prompt text","");` would make more sense for normal browser usage. The best approach might be to abstract the general "readline()" capability behind a batavia interface (much like stdout and stdin have been), with a default behavior that is "browser compliant". 

If possible, some light tests of the `input()` function would also be nice - even if it's just testing that a `readline()` testing shim is being activated.

pietertolsma:
@freakboy3742 sounds good! Working on it right now :) Might take some time because I've got my final exams atm as well. Keep on hacking!

pietertolsma:
@freakboy3742 in what class would I implement such an interface? Sorry for bothering you

freakboy3742:
I'd say either `batavia/batavia.js` (which is where the stdout/stderr definition is currently) or a utility method in `batavia/utils.js` would make sense.

freakboy3742:
@pietertolsma Did you close this intentionally? I'm not sure if you've got a deeper purpose for closing the PR, but if it's just because of the merge conflicts, you don't need to close this PR to fix those - you can correct them in a commit and then push, and the PR will be updated.

For what it's worth, the bits that aren't the merge conflict look good...

pietertolsma:
@freakboy3742 yeah I did! Thought it was unprofessional, will open it again. Still struggling with trying to understand the testing.

freakboy3742:
@pietertolsma No worries! When I merge, I'll collapse everything, so all the main repository will see is a single commit that contains the diff that is required.

As for testing: on reflection, this is going to be slightly complex to test, because you're going to need to inject some "test preparation" code into the virtual machine before running the Javascript. The test preparation code will inject an implementation of `batavia.input` that provides pre-canned input, rather than actually going to the keyboard and requesting input. If that sounds a bit daunting, don't worry about the test - test the new code manually and I can open a ticket to add a test as a project for someone else.

One last thing - for the sake of consistency, calling the configurable endpoint `batavia.stdin` rather than `batavia.input` would be preferable.

freakboy3742:
This was addressed in 1816445a.

PR #64, opened 2016-05-15
Implementation of operations on standard types - test_add_bool (issue 46)

winza83:
Attempt to implement test_add_bool for datatypes.test_str .  

Resolved and pass tests for:
test_add_bool
test_add_int


freakboy3742:
This patch is starting to look really good. A couple of notes:

* You don't need to commit the contents of `__pycache__` or `.egg_info` directories, or any `.pyc` file. These are compiled artefacts that will be automatically generated when you run the code, so they don't need to be version controlled.

* You also don't need to commit `batavia.js` and `batavia.min.js`. Again, these are compiled artefacts - but this time, they're the product of the build process. Version control only needs to contain the sources; anything that can be generated from source generally doesn't need to be in version control.

* You've inadvertently passed two more tests than you were targeting. There are two unexpected successes in the suite when it ran through CI:
  - test_add_str (tests.datatypes.test_int.BinaryIntOperationTests)`
  - test_add_str (tests.datatypes.test_bool.BinaryBoolOperationTests)
  So, you can remove these two lines as well.

* I really like the `datatype()` abstraction you've added - I've been considering a similar utility method. A minor preference thing would be to call it `isinstance(x, y)`, returning a boolean - that would make the method consistent with Python itself.

* It's generally a good idea to keep an eye on local coding conventions and adhere to them. For example, the existing code uses 4 spaces for indents, not tabs or 5 spaces. The existing code also uses "K&R" style braces i.e.:
  ```
if (condition) {
    ...
} else if (other condition) {
    ...
} else {
    ...
} 
   ```
  Your patch uses a combination of different styles. These style things don't have any effect on how the code executes, but it can make a difference to readability for others. It's a bit like cleaning your room - sure, it doesn't *strictly* matter if you've got cake crumbs all over the floor, but if you want to make a good impression on someone who is visiting, it's generally a good idea to vacuum before they get there. It's also one of those things where once you're used to living in a house with pristine floors, walking into a room where someone *isn't* a clean freak - it really stands out.

I know this probably seems like a mountain of stuff to fix, but please don't be disheartened. You're making really good progress - these might seem like really big things, but honestly, they're relatively minor. You're not far away from a commit, I promise!

winza83:
Thanks for the feedback @freakboy3742.  I'll fix these and will make another attempt.

freakboy3742:
This has been superseded by #69.

PR #66, opened 2016-05-18
Adding line to change into correct directory

FlaviaBastos:
I got the following error when running 'pip install -e .":
`Directory '.' is not installable. File 'setup.py' not found.`

Then I cd batavia, ran pip install again and it went through this time. I guess line was missing from this doc.


freakboy3742:
Thanks for the eagle eyes! Seems like no matter how much you look at a set of instructions, you always miss something obvious :-)

PR #67, opened 2016-05-18
Implement correctly printing round floats (e.g. 0.0) plus type() and abs()

wlmiller:
I set out to implement the builtin `abs()`, but realized that in order to get the test passing for floats I needed `print()` to work for floats and an implementation of `type()`.  I ended up creating a new type for floats so that, for example, `0.0` and `0` are distinct.

This required some minor tweaks to some existing methods but all tests should be passing.


wlmiller:
Strange - I added `@unittest.expectedFailure` to a test that fails on my machine, but that caused a build failure on Travis CI due to unexpected success.  I'll pull off the expected failure.

batavia/utils.js
```diff
@@ -324,6 +324,37 @@ batavia.core.tuple_iterator.prototype.__str__ = function() {
...
+    };
+    
+    Float.prototype.valueOf = function() {
+        return this.val;
+    }
+    
+    return Float;
+}();
+
+batavia.core.tuple_iterator.prototype.__str__ = function() {
```
freakboy3742: Is this a copy & paste error? Should this be Float, rather than tuple_iterator?

batavia/core/builtins.js
```diff
@@ -995,8 +1002,26 @@ batavia.builtins.tuple = function(args) {
     return new batavia.core.Tuple(args[0]);
 };
 
-batavia.builtins.type = function() {
-    throw new batavia.builtins.NotImplementedError("Builtin Batavia function 'type' not implemented");
+batavia.builtins.type = function(args) {
+    var type = 'NoneType'
```
freakboy3742: At the top of builtins.js, there's a template for builtin methods - they should take (args, kwargs) as arguments, and do some light validation of those arguments to make sure you're not passing in too many/not enough arguments. There's also a `__doc__`  attribute that should be added.

batavia/utils.js
```diff
@@ -412,30 +447,58 @@ batavia.operators = {
...
+            && ((typeof a.valueOf() == 'number' | typeof a == 'boolean')
+                && (typeof b.valueOf() == 'number' || typeof b == 'boolean'))) {
+            if (a instanceof batavia.core.Float || b instanceof batavia.core.Float) {
+                result = new batavia.core.Float(a*b);
+            } else {
+                result = a * b;
+            }
+        } else {
+            var aType = batavia.builtins.type([a]).split("'")[1];
+            var bType = batavia.builtins.type([b]).split("'")[1];
```
freakboy3742: This will absolutely work - but it involves a linear scan of the content, creating an array with three elements, then throwing away 2 of them. I suspect it would be faster if you just did a `slice(1,-1)` (since we know the quotes are the first and last elements) - that way there's just a single substring being created, based on known string indicies.

batavia/utils.js
```diff
@@ -412,30 +447,58 @@ batavia.operators = {
...
     },
     DIVIDE: function(a, b) {
-        return Math.floor(a / b);
+        if (a != null && b != null 
+            && ((typeof a.valueOf() == 'number' | typeof a == 'boolean')
+                && (typeof b.valueOf() == 'number' || typeof b == 'boolean'))) {
+            if (a instanceof batavia.core.Float || b instanceof batavia.core.Float) {
+                result = new batavia.core.Float(a*b);
+            } else {
+                result = a * b;
```
freakboy3742: Shouldn't this be `/`?

freakboy3742:
This is amazing - thanks for the contribution! 

I've added a couple of minor requests for fixes - a couple of minor bugs and consistency issues; shouldn't be too much effort to fix if you've been able to get this far :-)

batavia/utils.js
wlmiller: It's actually not quite that simple - a float for example has a type() of `<class  'float'>`.  But you're absolutely right it doesn't make sense to construct that string only to deconstruct it immediately, so I'll pull out the functionality for getting the type as a string (e.g. just `float`) into a separate utility method.

batavia/utils.js
```diff
@@ -791,3 +787,29 @@ batavia.make_callable = function(func) {
     fn.__python__ = true;
     return fn;
 };
+
+/*************************************************************************
+ * get the argument type as a string
+ *************************************************************************/
+ 
+ batavia.get_type = function(arg) {
```
wlmiller: Hopefully this is an okay place for this method - I didn't see a better spot but let me know and I'm happy to move it.

freakboy3742: The location looks as good as any. The only improvement I can think of would be to use an actual class rather than a string - but that's something we can clean up later.

freakboy3742: Oh - wait - this is to get a printable version of a type, rather than the "type" itself. In that case, there's probably something we can do about actually making the class itself have a `__str__` method... but that's a much bigger job, so feel free to ignore that for now. The only fix I'd suggest is to rename to `get_type_name` so it's clear that it *is* a string, not something that should be a type descriptor.

freakboy3742:
This looks good - the only lingering problem is the use of `*` in the `DIVIDE` implementation - I'm pretty sure that will give you another dozen or so test passes.

wlmiller:
Oops.  In the process of fixing that, I realized that it looks like `DIVIDE` isn't actually called - it's `TRUE_DIVIDE` - so I updated `TRUE_DIVIDE` (with the correct error message), added a `ZeroDivisionError` check, and also went ahead and implemented `FLOOR_DIVIDE` to get a few more test passes.

freakboy3742:
Fantastic - thanks for the contribution! If you've got a Twitter handle, I'd love to give you a shout out.

wlmiller:
Absolutely - it was fun to dig in. This is a cool project, and hopefully I'll be able to jump on something else soon.

I'm not particularly active on Twitter but I do have a handle: @nealmillr.

PR #74, opened 2016-06-02
Implement Polyglot error [Design Issue]

glasnt:
Some python internals have absolutely no meaning in the Batavia/JavaScript context

This PR sets up the framework for `id` being one of those things. 

@freakboy3742: your BDFN call is required.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ distribute-*
 /batavia.min.js
 /tests/temp
 env
+.eggs/
+.python-version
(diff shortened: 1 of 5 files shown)


phildini:
I look forward to having a design powwow with @freakboy3742 and @glasnt at PyCon sprints. My $0.02 is that, in order to support developers developing Python for Batavia as friction-free as possible, we should try to "fail open", ie: throw a warning in browser console but don't throw an exception.

glasnt:
As per design meeting, this is getting merged in. 

First pass, we'll make the things break hard. 

Second pass, if we want to later, we can easily find these decisions and change them accordingly.

PR #83, opened 2016-06-02
Updated contributing.rst to fix a minor formating issue.

BrianKayWalker:
A very subtle issue, Windows wasn't properly highlighted, just added another ~ and it took.


glasnt:
Nice catch!

Issue #26, opened 2016-04-02
Bytecode changes for dict in python 3.5 causes batavia.core.Dict to fail


splab:
With python 3.5 the bytecode for storing data to the dict has changed to:

```
  2           0 LOAD_CONST               1 ('a')
              3 LOAD_CONST               2 (1)
              6 LOAD_CONST               3 ('b')
              9 LOAD_CONST               4 (2)
             12 BUILD_MAP                2
             15 STORE_FAST               0 (lonk)
```

causing dicts to be empty.

To reproduce use (in the testbed):

``` javascript
foo = {'a': 1, 'b': 2}
print('bar', foo)
```

Expected result: bar 1
Actual result: bar

miki725:
so the fix is available on 3.5 branch. will try to get it running for both 3.4 and 3.5

slick666:
This issue for 3.5 has been addresses. issue in Python 3.6 are continues in #181

judy2k:
@freakboy3742 I think this issue has been accidentally re-opened. There's a separate issue for Python 3.6: #181

Zac-HD:
Was this fixed by #89?

freakboy3742:
Yup!

PR #86, opened 2016-06-02
Added BATAVIA_MAGIC which allows for Py3.4 and Py3.5 support

miki725:
The deal has to do with that fact that byte code has slightly changed in Python3.5. This change attempts to add support for batavia for both Py3.4 and Py3.5 by storing the byte code magic number in a window global hence allowing to check for it in the runloop.

The PR is related to https://github.com/pybee/batavia/issues/26 as well as https://github.com/pybee/batavia/issues/41.


miki725:
closing in favor of https://github.com/pybee/batavia/pull/89 since 3.5 branch is out of date

PR #88, opened 2016-06-02
Implement setattr on string

tfors:


tests/datatypes/test_str.py
```diff
@@ -4,7 +4,6 @@
 
 
 class StrTests(TranspileTestCase):
-    @unittest.expectedFailure
```
phildini: I think this test need to be updated? It reads that you _should_ be able to set arbitrary attrs on a `String` object. I'm not sure that's the case, but either the test or your code needs updating.

Thanks for the PR!

tfors: It looks like the build failures are a Travis CI issue.  I was able to successfully pass all tests locally using both 3.4.2 and 3.4.4 where Travis CI only passed on 3.4.4.  

The error on the 3.4.2 build had to do with trying to download Python 3.4.2.

PR #89, opened 2016-06-02
Added BATAVIA_MAGIC which allows for Py3.4 and Py3.5 support

miki725:
The deal has to do with that fact that byte code has slightly changed in Python3.5. This change attempts to add support for batavia for both Py3.4 and Py3.5 by storing the byte code magic number in a window global hence allowing to check for it in the runloop.

The PR is related to https://github.com/pybee/batavia/issues/26 as well as https://github.com/pybee/batavia/issues/41.

Follow up to https://github.com/pybee/batavia/pull/86 since 3.5 branch was out of date and its easier to merge the upstream changes to open new PR against master.

- [x] add test case for travis to run on Py3.4 and Py3.5
- [x] adjust `runAsJavaScript` to accept `bytes` which will it will assume is valid Python codecode and so will skip Python code compilation. this will allow to test the switch statement in JS within a single Python version
- [x] fixed for 3.5

```diff
- TypeError: list indices must be integers, not tuple
+ TypeError: list indices must be integers or slices, not tuple
?                                         ++++++++++
```


setup.py
```diff
@@ -10,7 +10,7 @@
 
 setup(
     name='batavia',
-    version='3.4.0.dev1',
+    version='3.5.0dev1',
```
miki725: this was the only significant merge conflict. Im not quite sure what is an expected value here now since this patch allows to run both Py3.4 and Py3.5

batavia/types/List.js
```diff
@@ -192,7 +192,11 @@ batavia.types.List = function() {
 
             return new List(Array.prototype.slice.call(this, start, stop));
         } else {
-            throw new batavia.builtins.TypeError("list indices must be integers, not " + batavia.type_name(index));
+            var msg = "list indices must be integers or slices, not ";
+            if (batavia.BATAVIA_MAGIC == batavia.BATAVIA_MAGIC_34) {
```
miki725: according to https://github.com/pybee/batavia/blob/master/tests/utils.py#L398 in some places we simply swap out the error message in Python during tests so that assertion passes. given that, Im not sure if this approach makes sense. this approach matches error message to of corresponding Python version whereas the other approach is a bit simpler.

@freakboy3742 thoughts?

PR #90, opened 2016-06-02
Updating contributing.rst as you are already in the batavia dir, you …

BrianKayWalker:
…don't want to go down again

You have earlier in the instructions moved down to the batavia directory you don't want to go down one more (there is one).


glasnt:
💯

PR #92, opened 2016-06-02
Trow new type error unsupported operand

netoxico:
Trow new type error unsupported operand divide string and bool

### DIFF
--- a/batavia/types/Str.js
+++ b/batavia/types/Str.js
@@ -217,7 +217,6 @@ String.prototype.__ifloordiv__ = function(other) {
 };
 
 String.prototype.__itruediv__ = function(other) {
-    //throw new batavia.builtins.TypeError("DDDDDDDD");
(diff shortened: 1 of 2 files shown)


batavia/types/Str.js
```diff
@@ -217,7 +217,9 @@ String.prototype.__ifloordiv__ = function(other) {
 };
 
 String.prototype.__itruediv__ = function(other) {
-    throw new batavia.builtins.NotImplementedError("String.__itruediv__ has not been implemented");
+    if (batavia.isinstance(other, batavia.types.Bool) ) {
+      throw new batavia.builtins.TypeError("unsupported operand type(s) for /=: 'str' and 'bool'");
+    }
```
phildini: Could you add a fall-through case here? Right now we just fall of the end if `other` isn't a `bool`.

PR #95, opened 2016-06-02
Throw a TypeError for in-place subtraction from a string

camdroid:


batavia/types/Str.js
```diff
@@ -233,7 +233,7 @@ String.prototype.__iadd__ = function(other) {
 };
 
 String.prototype.__isub__ = function(other) {
-    throw new batavia.builtins.NotImplementedError("String.__isub__ has not been implemented");
+    throw new batavia.builtins.TypeError("unsupported operand type(s) for -=: 'str' and '" + batavia.type_name(other) +"'")
```
phildini: Could you look at e916704fdb4a9c629b86bdb06324509ec3c099d2? I think the approach taken there is better, because it will be easier to change when we implement the "good" behavior.

camdroid: Um, I'm not sure I understand, what's the "good" behavior? Doesn't batavia.type_name() account for when the argument is null or NoneType? And why would the NotImplementedError still be left in?

phildini: Looking at e916704fdb4a9c629b86bdb06324509ec3c099d2, they've implemented a series of if statements to catch different types of "bad" input. This is great, because it means we can append to the work they've done or replace their `if` statement when we're ready to finish that function.

Since your work is doing similar things, it would be great if it followed a similar pattern. Following that pattern will make it easier to edit later, and be closer to the complete solution.

camdroid: Won't the behavior be the same for both branches of the if statement? If `other` is null or NoneType, then `batavia.type_name(other)` will return `"NoneType"`, and the result would be the same as if lines 225 and 230 were duplicates.

Why would we want to split it up into branches when all the branches would do the same thing?

PR #97, opened 2016-06-02
modified __pow__ so it raises a type error

sethMittag:


freakboy3742:
Almost perfect - just one tweak required. The commented out line isn't needed - it looks like it's part of an earlier draft or debugging process.

So - delete line 104 (the commented out work in progress line), commit, push, and you'll be done.

sethMittag:
I believe this is fixed now

PR #112, opened 2016-06-03
include type names in core types

schuyler1d:
silly patch just to get us __name__ on the core types.


phildini:
How does this normally get used? I'm trying to find the `__name__` of an int in a python 3 interpreter, and not able to see that property at all.

freakboy3742:
`__name__` isn't on the instances, but it is on the *type*. In the long term, there's some work to be done in getting the distinction between types, classes, and instances sorted out; but for now, putting it on the class as an attribute is better than nothing.

freakboy3742:
Also - looks like this has been included in @schuyler1d's other patch, which has already been merged.

PR #116, opened 2016-06-03
Limit the testsuite checking for error messages [Major design PR]

glasnt:
There are issues with the slight variations of error messages across different versions of Python (3.4.4 vs 3.4.2, etc)

One way to solve this is for the test suite to specifically limit the forcefulness of the exact matches of error messages. 

With this change, only error types are compared, not specific error messages as well

Now, this might be too generous. We might not be catching enough (`unexpected success` results from the test suite are var too numerous)

However, with this change, we should be able to target more Python versions without having to explicitly change error messages all over the place. We use 3.4.4 as the default for messages, but we don't enforce this in the test suite; however the end user experience should still be great. 

@freakboy3742 @phildini


phildini:
Since @glasnt will (I think) be in transit for the next ~24 hours, I'm going to pull this into my local copy and see if I can't get it to work.

glasnt:
I think there's just the removing of the unexpected successes to go (which
was my last commit)

I endeavour to get this merged before I fly away. (she says now)

rbtcollins:
Thats great. You might consider using a structured protocol from test suites too (e.g. subunit, tap, junitxml)? [Assuming I understand the intent here, came here from twitter, so sorry if I've totally misunderstood whats going on :))]

glasnt:
OK, a double more rounds and then a squash and this should be OK to go

I'm getting a bit confused when there's duplicate test names in the same file, but under different sections, hence the movement in unexpected success vs failure counts. There's also the test name duplications between datatype and builtins that means I think I've been removing the wrong thing once or twice

Since I don't think this is blocking anyone I'll keep at it and ping a core team member when it's done

glasnt:
@freakboy3742 @phildini It's all green! It's alive! IT'S ALIVE! 🍏 🚀

phildini:
I'm totally ready to 🚢 this, size of the change makes me want a once-over from @freakboy3742

freakboy3742:
PUSH THE BUTTON! :shipit:

PR #124, opened 2016-06-04
Issue 123 - return valid argument count

lrhewus:
Update len function error to return the correct number of arguments
given when greater than 1.


freakboy3742:
Thanks for the contribution - I don't know if you're at the PyCon sprints, but if you are, come find me, because I've got a challenge coin for you!

lrhewus:
I'll be there on Sunday!

PR #129, opened 2016-06-04
Implement set()

ashley-dunn:
Implement `set()`, fixed a bug in the Set constructor, fixed string representation of sets, and removed passing tests.


freakboy3742:
Added in 59dedc8

PR #130, opened 2016-06-04
updated internals, so it is more clear now that tests may take up to …

Brigette-Eckert:
…1.5hrs to run

### DIFF
--- a/docs/internals/contributing.rst
+++ b/docs/internals/contributing.rst
@@ -189,7 +189,8 @@ You're now ready to run the test suite! From the batavia-dev/batavia directory T
 
     $ python setup.py test
 
-This will take about 5 minutes on most modern PCs/laptops, and will generate around 4000 lines of console output - one line for each test that is executed. Each line will tell you the pass/fail status of each test - e.g.,::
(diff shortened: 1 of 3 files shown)


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
phildini: Since we've added `.idea/` to the gitignore, it's strange that this is showing up? Could we remove it?

docs/internals/contributing.rst
```diff
@@ -198,7 +199,7 @@ This indicates that tests have passed (``ok``), or have failed in an expected
 way (``expected failure``). These outcomes are what you expect to see. If you
 see any lines that end ``FAIL``, ``ERROR``, or ``unexpected success``, then
 you've found a problem. If this happens, at the end of the test run, you’ll
-also see a summary of the cause of those problems.
+also see a summary of the cause of those problems. If you see "ERROR" press cntrl-c or cmd-c to quit the tests to debug.
```
phildini: ctrl is preferred over cntrl

PR #138, opened 2016-06-04
Update str.prototype.ge, str.prototype.le, str.prototype.gt

Brigette-Eckert:
contains push for str.le, st.gt and ge.


freakboy3742:
Merged in 73061e0

PR #145, opened 2016-06-05
updated contributing doc, to make the fact that tests may take over a…

Brigette-Eckert:
…n hour clear, and how to stop test


Brigette-Eckert:
Updated documentation so it is clear that tests may take upwards of 1.5 hours.  Added how to stop tests in case of error.

PR #146, opened 2016-06-05
str_inplace_shift_errors

jlebzelter:
* fix type in error strings
* Add correct error for ilshift, irshift


jlebzelter:
This test failed because it did not finish.  How to I trigger it to run again without pushing a change?

PR #147, opened 2016-06-05
Added handling of compatable data types for int_le

lrhewus:


phildini:
I think this looks really good, I'll see if we can get tests sorted out.

PR #148, opened 2016-06-05
added iterable check for all()

LawlietBlack:
other tests dependent on implementation of __iter__


LawlietBlack:
Tests timed out

PR #150, opened 2016-06-05
dict comparison handling, to include python-style ==

ivanlyon:
Filled out dict() comparison functions.  Handled == as python styled equality operator between dicts.


batavia/types/Dict.js
```diff
@@ -52,19 +52,30 @@ batavia.types.Dict = function() {
      **************************************************/
 
     Dict.prototype.__lt__ = function(other) {
-        return this.valueOf() < other;
+        throw new batavia.builtins.TypeError("unorderable types: dict() < dict()");
```
phildini: What if `other` is an int or string or...?

This function should check for types and return the correct error message.

Alternatively, if you'd rather fill out all the types right now, throw the `NotImplemented` error if the type isn't `dict`.

batavia/types/Dict.js
```diff
@@ -52,19 +52,30 @@ batavia.types.Dict = function() {
...
     };
 
     Dict.prototype.__le__ = function(other) {
-        return this.valueOf() <= other;
+        throw new batavia.builtins.TypeError("unorderable types: dict() <= dict()");
```
phildini: see above.

batavia/types/Dict.js
```diff
@@ -52,19 +52,30 @@ batavia.types.Dict = function() {
...
     };
 
     Dict.prototype.__eq__ = function(other) {
-        return this.valueOf() == other;
+        if (batavia.isinstance(other, [batavia.types.NoneType]) == true) return false;
```
phildini: See above about what should happen if the type is not a dict or None.

throwing the `NotImplemented` for the other types is fine here too.

phildini:
Thanks for submitting this! It's a good improvement. I'd like a few more cases covered, either by correct behavior or `NotImplemented` before accepting.

batavia/types/Dict.js
ivanlyon: robert in15 dict_window

Pythonic "==" appears to throw no errors when dict is involved.  I can change it regardless, if I receive confirmation.

batavia/types/Dict.js
ivanlyon: Good catch, changed in my latest commit.

batavia/types/Dict.js
ivanlyon: Good catch, changed in my latest commit.

freakboy3742:
Unfortunately, it appears that due to the confusion during the sprints, we've merged a different version of the comparison operators for dict, contributed by another user. Apologies for the time waste!

PR #154, opened 2016-06-06
Test dict types

thewus:
It's the little things that count.


glasnt:
Ah yes. That would help. 

Just kicking off the failed build again (timeout), then this should be good to go

thewus:
Travis-CI has been temperamental.
[#149 could use another shot also](https://travis-ci.org/pybee/batavia/builds/135464041)

PR #158, opened 2016-06-10
Remove `__idiv__` placeholders from type definitions.

phildini:
`__idiv__` does not exist in python3.
https://docs.python.org/3/library/operator.html


sethMittag:
Hey @phildini when I was working on __truediv__ for set, I noticed that __div__ was not in python3 yet was included.  I don't know if this is the appropriate place to bring this up but I saw that you were removing __idiv__ so I thought I would say something.

PR #166, opened 2016-06-19
Test suite speedups and normalization

freakboy3742:


glasnt:
Awesome! 👏 
The tests are now under 7 minutes each, down from 15-minutes-or-timing-out during the sprints at the start of the month! 🎉

PR #176, opened 2016-07-21
Added test_add_str and updated NotImplementedError to TypeError for NotImplementedType

jcronyn:
Worked on test_add_str, who's default implementation was correct with the exception of not throwing the correct error for NotImplementedType __add__ and __iadd__.


jcronyn:
Don't merge this - just realized I made a mistake in the implementation.

PR #178, opened 2016-07-21
Implement time.time() and time.sleep(), with tests.

dmpayton:
As noted in the title, this PR implements `time.time()` and `time.sleep()`.

The test for `time.time()` is currently marked as an expected failure because ... how do you test it? You can't compare the return value because every subsequent call returns a different value. You may be able to mock it out, but you'd have to do this in both JS and Python.

CPython seems to sort of punt it, but I'm not entirely sure how that test[0] is supposed to work. In any case, `time.get_clock_info()` isn't implemented in batavia, soooo...

[0] https://github.com/python/cpython/blob/master/Lib/test/test_time.py#L60

### DIFF
--- a/batavia/modules/time.js
+++ b/batavia/modules/time.js
@@ -2,5 +2,23 @@
 batavia.modules.time = {
     clock: function() {
         return new Date().getTime();
+    },
+
+    time: function() {
+        // JS operates in milliseconds, Python in seconds, so divide by 1000
+        return new Date().getTime() / 1000;
+    },
+
+    sleep: function(secs) {
+        if (secs < 0) {
+            throw new batavia.builtins.ValueError('sleep length must be non-negative')
+        }
+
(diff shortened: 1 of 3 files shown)


batavia/modules/time.js
```diff
@@ -2,5 +2,23 @@
 batavia.modules.time = {
     clock: function() {
         return new Date().getTime();
+    },
+
+    time: function() {
+        // JS operates in milliseconds, Python in seconds, so divide by 1000
+        return new Date().getTime() / 1000;
```
phildini: Should we force this to be a Float type or will it do the right thing?

dmpayton:
I've updated `time.time()` to return an explicit float.

The test for it is still marked as an expected failure due to (I think) an issue with `isinstance` where `isinstance(time.time(), float)` is returning `false`.

Issue #46, opened 2016-04-13
Complete implementation of operations on standard types


freakboy3742:
Javascript has a specific set of allowed operator and operations. Python has a set of operators and operations as well. However, the two don't match exactly. For example, Javascript behavior for adding/multiplying with Strings leads to behaviour that is quite distinct to the behavior provided by Python.

In order to replicate Python behavior in Javascript, Python's logic for all the basic operations needs to be implemented in the Batavia support library.

The test suite contains around 10000 tests in the that are currently marked as "expected failures". These reflect an attempt to do "every operation between every base data type". The task: pick a data type, and write the implementation of one of the math operation or comparison operators.

The tests will pass (and become unexpected successes) when the output of Python code doing that operation is the same when run through CPython and Batavia - and I mean byte-identical, down to the text and punctuation of the error message if appropriate.

If you want to see an example of what is involved, check out the implementation of add for integers. If you have an int, Python will allows you to add an int or a float to it; all other types raise a TypeError. The list of operations on int that still need to be implemented can be found [here](https://github.com/pybee/batavia/blob/master/tests/datatypes/test_int.py#L36); if you add a new operation, delete the line that corresponds to that test, and the test will report as a pass, rather than an expected fail.

betojulio:
I will work on test_multiply_float for String

winza83:
I'm still working on test_add_bool.

betojulio:
working on test_add_list

betojulio:
working on test_add_none for string

betojulio:
working on test_and_none for string

betojulio:
working on test_multiply_none for string

betojulio:
working on test_unary_positive for string

vyyyy:
working on test_subtract_str for String

FlaviaBastos:
Working on test_floor_divide_str for string

FlaviaBastos:
Ops. I guess test_floor_divide_str got resolved by @wlmiller on issue #67 
I will pick something else to work on.

wlmiller:
I'm sorry about that - I should have more carefully checked what was already being worked on.

For what it's worth, tackling these operations one at a time for each pair probably isn't the best way to go about it.  For example, if you resolve `test_floor_divide_str` for string, you're very likely to also resolve `floor_divide` tests for many other combinations of types (in particular combinations which can't be `floor_divide`d, e.g. string // tuple, list // dict, etc.).

[...]

PR #185, opened 2016-07-22
implements slice step for List and Str types __getitem__. refs #46

joslarson:
implements slice step for List and Str types __getitem__. refs #46

### DIFF
--- a/batavia/types/Slice.js
+++ b/batavia/types/Slice.js
@@ -4,7 +4,7 @@
 
 batavia.types.Slice = function() {
     function Slice(kwargs) {
-        this.start = kwargs.start || 0;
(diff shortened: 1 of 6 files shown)


batavia/types/Slice.js
```diff
@@ -4,7 +4,7 @@
 
 batavia.types.Slice = function() {
     function Slice(kwargs) {
-        this.start = kwargs.start || 0;
+        this.start = kwargs.start;
```
joslarson: Old code assumed a default of `0` for a slices start value, but when you use a negative step the default value for start is actually the length of the list/str/etc. So it makes more sense to allow a default of null (same as stop) and to handle setting the default in the slice implementation.

PR #195, opened 2016-07-22
fixes Str slice step tests to use strings, fixes string reverse logic…

joslarson:
fixes Str slice step tests to use strings, fixes string reverse logic in str \__getitem\__. ref #46


glasnt:
Hi @joslarson. Travis is failing because of another unexpected success, like you've [cleaned up before](https://github.com/pybee/batavia/commit/7057f89818812636cafb96d7bd25f358791e2c9e). Make that change and Travis will be happy.

joslarson:
@glasnt I got around to fixing the unexpected success, but something went wrong with 3.4 build before it even started on Travis (3.5 worked fine). Are you able to force rerun the tests to see if they pass? I did merge master this time before pushing, not sure if that could be the problem.

PR #201, opened 2016-07-23
Typo

Quantum238:
Very minor typo.  On a related note, the code-block:: doscom doesn't seem to show up anywhere - not in the github version of these docs or the hosted readthedocs.  Similarly, the code block for anaconda doesn't appear (the page says For those using anaconda: and then there's nothing).  I don't quite know how to fix that


freakboy3742:
Thanks for the patch. I *think* the issue is that you need to have a blank line between the `.. code block::` bit, and the actual code block. If you run sphinx over the docs and check the output, that should confirm it.

PR #202, opened 2016-07-24
Adjustments to the test running code for Windows compatibility

Quantum238:


freakboy3742:
Fantastic - thanks for this! @dcstlouis was having problems with this during the sprints, and we were able to narrow it down to an encoding issue, but I didn't know about the PYTHONIOENCODING thing.

PR #203, opened 2016-07-24
Implement complex

timb07:
Partial implementation of complex type (and some other tweaks); helps to address #46 and #47.


glasnt:
Wowzers! Nice work, @timb07!

If you're happy with your functionality in this PR, I can squash-merge for you. I just want to make sure you didn't have any other cleaning up stuff you wanted to get in before I do so

timb07:
It's mostly feature complete, the two main exceptions being creating a complex object from complex "real" and "imaginary" components, and pow(). If anyone desperately needs those for Batavia to be useful, I'll be a very surprised rabbit.

There are a few changes that aren't related only to the implementation of complex, such as the "Confused END_FINALLY" change. I don't know whether those are better squashed or not, so I'll defer to your opinion.

PR #204, opened 2016-07-24
Implement builtin 'reversed' to return a reversed iterable

Vector919:
I understand this isn't 100% analogous to the behavior of reversed in Python 3.4, but I didn't see an existing concept of generators yet.

### DIFF
--- a/batavia/core/builtins.js
+++ b/batavia/core/builtins.js
@@ -870,7 +870,7 @@ batavia.builtins.reversed = function(args, kwargs) {
     if (batavia.isinstance(iterable, [batavia.types.List, batavia.types.Tuple])) {
         var new_iterable = iterable.slice(0);
         new_iterable.reverse();
-        return new_iterable;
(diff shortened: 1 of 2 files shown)


freakboy3742:
@Vector919 This looks great; the only thing missing from the patch are some tests for the behavior of `reversed`. There's a placeholder for these tests already - [the `ReversedTests` class in `test_reversed.py`](https://github.com/pybee/batavia/blob/master/tests/builtins/test_reversed.py#L4); you just need to add some tests that validate that the method actually returns a reversed list.

Vector919:
@freakboy3742 I've added a test asserting that the code execution is the same, so it looks like I need to wait for this to rebuild

freakboy3742:
Fantastic - Thanks for the contribution! If you wanted to bulk out those tests - for example, validating that an empty list can be reversed, or that the reversed list doesn't touch the original list, that would be great. Or, feel free to move onto another method!

PR #209, opened 2016-07-28
Added tests for the 'all' builtin

Vector919:
It seems like there's missing tests for a lot of the builtins. If this PR works, most of the rest of them can be easily tested

### DIFF
--- a/batavia/core/builtins.js
+++ b/batavia/core/builtins.js
@@ -168,6 +168,9 @@ batavia.builtins.abs = function(args, kwargs) {
 batavia.builtins.abs.__doc__ = 'abs(number) -> number\n\nReturn the absolute value of the argument.';
 
 batavia.builtins.all = function(args, kwargs) {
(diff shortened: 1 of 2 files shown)


freakboy3742:
Agreed that many of the builtins need a lot more testing. The tests you've already submitted prove that - you've found a bug in the output when iterating over non-lists. If you want to spend some time working up patches like this one, it would definitely be appreciated.

The two tests you've presented here are great. However, the second one (`test_all_typeerror`) needs to be marked as an expected failure (or the implementation corrected). The first test would also benefit from exercising a couple of other use cases, e.g.:
* An empty list
* A list that where `all()` evaluates `True`
* A generator expression
* A list that only contains False-ish elements

Vector919:
@freakboy3742 I've added the test cases that you suggested, as well as corrected the implementation of `.all()` to match what actually happens in python3

PR #221, opened 2016-08-15
Ready To Merge: implement bytes, Python3's str, bytearray

candeira:
I'm only opening this PR to produce work-in-progress reports of working with bytes and bytearrays. 

Done so far:

[x] Implement `bytes()` on JS `buffer.Buffer`
[x] `bytes()` initialisation from: no args, int, book, bytes, list, str (no encoding),  None, dict, list, class, float, NotImplemented, set, tuple, slice.
[x] Bytes methods implemented: `__len__`, `__str__`
[x] Make b'foo' == bytes([102, 111, 111]) (before, b'foo' would build a native JS string)
[x] Add small_ints support to BuiltinFunctionTestCase
[x] `bytes()` initialisation from: string + encoding
[x] `Bytes` methods: `__eq__`, `__gt__`, etc.
[x] `bytes()` initialisation from other iterables (frozenset, range, after implementing them)
[x] `str.encode(encoding)`
[x] `bytes.decode(encoding)`

To Do:

[ ] `str(bytestring, encoding, errors)`, so Python3 `str()` (the one now is Python2's)
[ ] add `kwargs` support to `bytes()` and `str()`
[ ] `int.to_bytes()`

I'll only do utf-8, ascii and latin1 encodings for now. Everything else is half as important/urgent.

After that, I'll leave the low-hanging fruit methods for newcomers to pick up, and move on to implement `bytestring()` and `memoryview` interface in general.


candeira:
The plan is to vendor Feross Aboukhadijeh's MIT-licensed https://github.com/feross/buffer, which re-implements Node's Buffer API with fast paths for ArrayBuffer on modern browsers, while having software fallbacks as far as IE6.

This also gives us full interoperability with JS code depending on buffers.

Note:  to confuse JS/NOde buffers with Python's own buffer (slice or view over another object). That will be a separate implementation.

candeira:
It appears that JS/Node Buffers are immutable, so `bytearray()` is going to take more planning. I'm just work on `bytes()` for now.

candeira:
Got sidetracked fixing Bool's __int__ and its __str__ representation. Still Work in Progress.

candeira:
Progress report: I'm still working on bytes, but the buffer JS dependency was badly vendored, so I had to fix vendoring first: https://github.com/pybee/batavia/pull/302

Hopefully soon.

candeira:
Code is now working nicely up to equvalence between b'foo' and bytes('foo', encoding=ascii). I'm now working on passing the tests given Batavia's existing fixture data for sequences (lists, sets, tuples) and next I'll move onto initialising from strings.

candeira:
This is kind of a "how long is a bit of string" thing, so I've decided to keep a To-do here in the pull request:

Done so far:

[x] Implement `bytes()` on JS `buffer.Buffer`
[x] `bytes()` initialisation from: no args, int, book, bytes, list, str (no encoding),  None, dict
[x] Bytes methods implemented: __len__, __str__
[x] Make b'foo' == bytes([102, 111, 111]) (before, b'foo' would build a native JS string)

To Do:
[ ] Bytes methods: __eq__, __gt__
[ ] `bytes()` initialisation from: string + encoding, other iterables (tuple, set, frozenset, range)
[ ] `str.encode(encoding)`
[ ] `bytes.decode(encoding)`
[ ] `int.to_bytes()`

I'll only do utf-8, ascii and base64 encodings for now. Everything else is half as important/urgent.

The problem with passing random iterables to the `bytes()` builtin is that something seems to unbox the `batavia.iter_for_each`-yielded items. Then, as JS `boolean` doesn't behave as an `int` like a Python `bool` does, the tests don't pass. But it seems to be an issue somewhere else, not in the `bytes()` implementation. I'll see if I can open a separate PR to fix this issue.

After that, I'll leave the low-hanging fruit methods for newcomers to pick up, and move on to implement `bytestring()` and `memoryview` interface in general.

candeira:
I think this is ready to go as soon as tests pass. Still some work to do, but what's done is usable already.

 Note:    

        - bytes(x:int) gives implementation-dependent errors for huge values,
          we use small_ints=True to skip testing with those values rather
          than cargo-culting the exact same exceptions. Examples:
    
          >>> f = bytes
          >>> x = -18446744073709551617
          >>> f(x)
        - <class 'ValueError'> : negative count
        + <class 'OverflowError'> : cannot fit 'int' into an index-sized integer
    
          >>> f = bytes
          >>> x = 1361129467683753853853498429727072845824
          >>> f(x)
        - <class 'OverflowError'> : byte string is too large
        + <class 'OverflowError'> : cannot fit 'int' into an index-sized integer

Next Steps:

        - str() builtin initialisation from bytes object.
        - bytes.decode(encoding) -> str
        - int.to_bytes(endianness) -> bytes
        - bytearray()
        - memoryview protocol has a lower priority,
          but maybe we should consider implementing
          Bytearray via memoryview.

tests/builtins/test_bytes.py
```diff
@@ -7,24 +7,21 @@ class BytesTests(TranspileTestCase):
...
+
+    def test_set(self):
+        self.assertCodeExecution("""
+            x = {66}
+            print(bytes(x))
+            """)
+
+    def test_frozenset(self):
+        self.assertCodeExecution("""
+            x = frozenset((66))
```
freakboy3742: Are you sure this example is testing what you think it is? Doesn't it need a comma after the 66 to make this a "set of a tuple"? Without the comma, it should raise an error -- which it may well do on both platforms, making the test pass, but that's not what I think you *think* you're testing.

batavia/core/builtins.js
```diff
@@ -1340,7 +1443,7 @@ batavia.builtins.vars.__doc__ = 'vars([object]) -> dictionary\n\nWithout argumen
 
 batavia.builtins.zip = function(args, undefined) {
     if (args === undefined) {
-        return [];
+        return new batavia.types.List;
```
candeira: Well spotted.

batavia/core/builtins.js
```diff
@@ -560,11 +663,11 @@ batavia.builtins.exec = function() {
 batavia.builtins.exec.__doc__ = 'exec(object[, globals[, locals]])\n\nRead and execute code from an object, which can be a string or a code\nobject.\nThe globals and locals are dictionaries, defaulting to the current\nglobals and locals.  If only globals is given, locals defaults to it.';
 
 batavia.builtins.filter = function(args, kwargs) {
-    if (arguments.length != 2) {
+    if (args.length != 2) {
         throw new batavia.builtins.BataviaError('Batavia calling convention not used.');
     }
     if (kwargs && Object.keys(kwargs).length > 0) {
-        throw new batavia.builtins.TypeError("filter() doesn't accept keyword arguments");
+        throw new batavia.builtins.xTypeError("filter() doesn't accept keyword arguments");
```
candeira: Erm.. xomeone ix using hix editor incorrectly...

tests/builtins/test_bytes.py
candeira: Thanks for this too. I've now fixed it by making a substitution that allows TypeErrors to come in any order for any type. Please see here: https://github.com/pybee/batavia/pull/221/files#diff-e236243f137e2c49514a1debf1b6765cR10

glasnt:
@candeira This PR now has merge conflicts. If you rebase, then CI should confirm that all is well, and we can merificate :)

candeira:
Merginating...

candeira:
Yay, thanks!

PR #222, opened 2016-08-15
Builtin pow function support for large arguments

swenson:
This helps with https://github.com/pybee/batavia/issues/47, as I am a first-timer!

There are still some issues with very large arguments, but I can file a separate bug for that, assuming it is not already known.

> Give me your math, your floor,
> Your huddled masses yearning to round free


swenson:
Hey look tests pass 🎉

PR #229, opened 2016-08-15
Adding styles to the Batavia demo page

### LINKED ISSUE #218: Create better styles for Batavia demo page
We can add some styles in the same style as the Batavia docs page, here: http://pybee.org/contributing/what/

We can add some font styles and other pretty changes to give a better first impression.

JasonThomasData:
This should fix #218, but whether's that's true comes down to taste as these are mostly style changes.

This PR actually turned into a bit of a ramble... sorry.

The Batavia demo page now has some styles that make it look a little nicer.
It has most styles in a separate file now since that made sense to me.

There's a line to explain what the provided example scripts are doing.

This has the Cutive font locally, since that's one fewer request.

It also has the CodeMirror license this time, which should have been included in the CodeMirror/syntax highlighting PR.


JasonThomasData:
@freakboy3742 I thought that final commit might fix the unicode error, but that didn't work :( More thought required obviously.

glasnt:
@JasonThomasData The unicode error is unrelated to your changes.

PR #235, opened 2016-08-16
Add builtin ascii function

MarkusH:


glasnt:
@MarkusH emoji tests! :tada:

... that break 😢 

the relevant output from CI: 

```
  
  >>> x = "🐍🐍🐝🐝"
  >>> y = slice(5)
  >>> x[y]
- 🐍🐍?
+ 🐍🐍🐝🐝
  
  >>> x = "🐍🐍🐝🐝"
- >>> y = slice(2, 7)
- >>> x[y]
- 🐍🐝?
- 
- >>> x = "🐍🐍🐝🐝"
- >>> y = slice(2, 7, 2)
?                   ---
+ >>> y = slice(2, 7)
  >>> x[y]
- ???
+ 🐝🐝
  
  >>> x = "🐍🐍🐝🐝"
- >>> y = slice(7, 2, -1)
?                   ----
+ >>> y = slice(2, 7, 2)
?               +++
  >>> x[y]
- ?🐝🐍
+ 🐝
  
  >>> x = "🐍🐍🐝🐝"
- >>> y = slice(7, 2, -2)
?                      ^
+ >>> y = slice(7, 2, -1)
?                      ^
  >>> x[y]
- ???
+ 🐝
+ 
+ >>> x = "🐍🐍🐝🐝"
+ >>> y = slice(7, 2, -2)
+ >>> x[y]
+ 
  
```

I think there's just a small little thing in the code that isn't working, but I'll let you dive into it.

MarkusH:
Noooooooo, what did I get myself into: https://mathiasbynens.be/notes/javascript-unicode

There's a DDN (Design Decision Needed) how to deal with UTF-16 / UCS-2 characters. Do we want to embed some 3rd party library for encoding handling, have something slow but working, ...

MarkusH:
Essentially, according to above link, there are 4 options how to count the length of characters:

```javascript
var regexAstralSymbols = /[\uD800-\uDBFF][\uDC00-\uDFFF]/g;

function countSymbols(string) {
	return string
		// Replace every surrogate pair with a BMP symbol.
		.replace(regexAstralSymbols, '_')
		// …and *then* get the length.
		.length;
}

function countSymbols(string) {
	return punycode.ucs2.decode(string).length;
}

function countSymbols(string) {
	return Array.from(string).length;
}

function countSymbols(string) {
	return [...string].length;
}
```

The last 2 are not supported on PhantomJS 2.1.1, the second on is a 3rd party package and the first one is sloooowww.

And that's only to start with.

glasnt:
@MarkusH so we do have the start of vendoring stuff in https://github.com/pybee/batavia/pull/230, but I want to know what/why `Array.from(string)` fails in PhantomJS 2.1.1 

Also, yes, emoji are hard, didn't you know? 🙃

glasnt:
Further: phantomjs doesn't support the new Array stuff in ES6 
https://kangax.github.io/compat-table/es6/
But we could polyfill it https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from

candeira:
Hi, @MarkusH, I'm adding the `byte` builtin by delegating to the Buffer library, which i've already put in `batavia/vendor/vendored.js`. Buffer already knows how to decode/serialise JS strings, so your `ascii` can depend on `bytes` for decoding, or maybe it can use Buffer directly. 

Here's the pull request where I'm working on this: https://github.com/pybee/batavia/pull/221

In either case, the big advantage of Buffer is that it knows to use fast paths like ArrayBuffer on modern browsers, and still works on older browsers.

candeira:
Hi, @MarkusH again. I've already implemented bytes('foo', 'ascii') and 'foo'.encode('ascii') in https://github.com/pybee/batavia/pull/221. The batavia implementation uses buffer's implementation of `Buffer.from(string [, encoding])`.

Maybe we should make the 'ascii' builtin just use the same path.

freakboy3742:
Superseded by #360.

PR #237, opened 2016-08-21
We don't need to install phantomjs ourselves on CircleCI

swenson:
Should save a few seconds per test, and won't need to download from S3. :)


glasnt:
Sounds good to me!

swenson:
Thanks!

PR #243, opened 2016-08-26
Check for some more tuple edge cases; implement tuple comparisons fully

swenson:
I added the empty tuple and other annoying edge cases to shake out some tuple-handling bugs, and then fixed those.


swenson:
Thanks!

PR #244, opened 2016-08-27
Keep a queue full of Python interpreters

swenson:
This should hopefully speed up tests by minimizing us waiting on CPython initialization.

If this works we can do the same thing for PhantomJS.


swenson:
Ehhhh. Nevermind. This doesn't seem to save us much on CircleCI, sadly, because we are limited to 1 core and it limits the number of subprocesses. :(

PR #250, opened 2016-09-03
Exception handler should clear the last exception

swenson:
If a block of code threw an exception, caught it, and then threw another exception, the first exception would be thrown twice. So, we clear the `last_exception` in between.

We also take into account that `object()` is not implemented yet to fix some method resolution order weirdness, which fixes a bunch of expected failures.


glasnt:
✨ ✨ ✨ ✨ ✨ ✨

swenson:
Thanks!

PR #256, opened 2016-09-05
Add a basic math module that wraps around existing JS Math functions

swenson:
Many Python `math` functions don't have exact analogues in JS `Math`, but we do basic wrappers around the ones that do. Others aren't universally supported, or just don't exist, and those we will have to write ourselves. In at least the case of `exp`, we have differences in printing floats that causes an error.

I also took this opportunity to fix `time.clock` to return a proper `float`, and be a counter that starts `0.0` (similar to, but not quite the same as, CPython).


swenson:
Hold up. Not sure why I didn't do something similar to the other builtin tests, so I'm going to redo these tests using those as a template to test all types, sample data, etc.

swenson:
Okay. Tests rewritten to be in the more consistent style. Ready for review.

glasnt:
@swenson I'm seeing that this failed due to a timeout, but it looked like you've kicked it off again.

swenson:
Yes. It failed the first time due to a "CircleCI error", and the second it timed out (though that same test worked fine when I ran it locally). Trying to see what's up.

swenson:
Tests pass now!

glasnt:
I think this looks good, and yay for Math and such, but I'm actually not sure as to what level the batavia-based implementation needs to do. All the standard library? Math, time, space, etc? 

I defer to the BDFN @freakboy3742

batavia/modules/math.js
```diff
@@ -0,0 +1,201 @@
...
+    // TODO: this isn't supported in PhantomJS
+    acosh: function(x) {
+        return new batavia.types.Float(Math.acosh(x.__float__().val));
+    },
+
+    asin: function(x) {
+        return new batavia.types.Float(Math.asin(x.__float__().val));
+    },
+
+    // TODO: this isn't supported in PhantomJS
```
freakboy3742: Is there a polyfill we can use to plug the gap here (and for the other hyperbolic arc trig functions) ?

batavia/modules/math.js
```diff
@@ -0,0 +1,201 @@
...
+    log: function(x, base) {
+        if (base == null) {
+            return new batavia.types.Float(Math.log(x.__float__().val));
+        }
+        return new batavia.types.Float(Math.log(x.__float__().val) / Math.log(base.__float__().val));
+    },
+
+    // TODO: this isn't supported in PhantomJS
+    log10: function(x) {
+        return batavia.modules.math.log10(x);
```
freakboy3742: Am I misreading this, or is this a recursive call? Shouldn't this be `batavia.modules.math.log(x)`

batavia/modules/time.js
```diff
@@ -1,7 +1,9 @@
 
 batavia.modules.time = {
+    _startTime: new Date().getTime(),
     clock: function() {
-        return new Date().getTime();
+        // TODO: this does something a little different in CPython.
```
freakboy3742: Can you elaborate on what "a little different" means here? I know this is going to be one of those functions where some discrepancy from CPython is to be expected, but it would be good to document exactly what that entails.

tests/modules/test_math.py
```diff
@@ -0,0 +1,316 @@
...
+        'test_math_pow_slice_frozenset',
+        'test_math_pow_str_frozenset',
+        'test_math_pow_tuple_frozenset',
+
+        'test_math_sin_bytearray',
+        'test_math_sin_bytes',
+        'test_math_sin_class',
+        'test_math_sin_complex',
+        'test_math_sin_dict',
+        'test_math_sin_float', # only fails for formatting reasons
```
freakboy3742: If this is a predictable formatting failure is predictable, you can put the float normalisation into the list [here](https://github.com/pybee/batavia/blob/master/tests/utils.py#L797) - this list compensates for known formatting discrepancies (substituting `value` for `key` in that dictionary at a text-content level)

batavia/modules/math.js
swenson: Ha! Oops. :)

batavia/modules/math.js
swenson: Looks there is one at https://github.com/MaxArt2501/es6-math -- I will add that.

batavia/modules/time.js
swenson: Reading through the docs for `time.clock`, it is allowed to return either the CPU time or the real time elapsed. So, this is technically within the specs.

I'll delete the comment.

tests/modules/test_math.py
swenson: Ah, good  to know, will do.

freakboy3742:
This looks really good; Some minor comments inline. 

The polyfill thing doesn't have to be tackled before we merge this - I'm just raising it as a question worth some exploration at some point. 

The only general comment I'd make is that it would be nice to have a `.__doc__` for each of these methods as well.

swenson:
Addressed the inline comments and added the polyfill. The polyfill is not as accurate as the CPython implementations, sadly.

I can add the `__doc__` for each in a follow-up PR. (I also have a PR in progress to do type checking on the methods to eliminate a lot of the expected failures of these methods.)

Makefile
```diff
@@ -31,6 +31,7 @@ ALL_FILES=\
     batavia/modules/dom.js \
     batavia/modules/marshal.js \
     batavia/modules/inspect.js \
+		batavia/modules/math.js \
```
phildini: actually weird tabbing or weird rendering error?

swenson: Weird tabbing -- now fixed. Thanks!

swenson:
Thanks!

PR #257, opened 2016-09-06
Add support for EXTENDED_ARG

swenson:
Cherry picked from an upcoming `math` change that generates a very large function, and requires this (and tested there).

### DIFF
--- a/batavia/modules/dis.js
+++ b/batavia/modules/dis.js
@@ -84,6 +84,7 @@ batavia.modules.dis = {
         batavia.modules.dis.def_op('DUP_TOP_TWO', 5);
 
         batavia.modules.dis.def_op('NOP', 9);
+        batavia.modules.dis.NOP = 9;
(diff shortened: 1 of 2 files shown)


batavia/modules/dis.js
```diff
@@ -84,6 +84,7 @@ batavia.modules.dis = {
         batavia.modules.dis.def_op('DUP_TOP_TWO', 5);
 
         batavia.modules.dis.def_op('NOP', 9);
+        batavia.modules.dis.NOP = 9;
```
phildini: Why are both the assignment and `def_op` needed? This is probably more for my education than an actual concern.

swenson: We need the constant `9` accessible as well, so that we can check if an opcode coming in is equal to this. The `def_op` function stores it in an object, which is a little more annoying to get at.

We do a similar thing elsewhere (in particular, `EXTENDED_ARG` at the bottom).

swenson:
Thanks!

PR #259, opened 2016-09-08
Implement a bunch more of the math library

swenson:
Adds the `math.copysign`, `math.degrees`, `math.radians`, `math.erf`, `math.erfc`, `math.fabs`, `math.factorial`, `math.fmod`, `math.frexp` functions.

`erf` and `erfc` are nearly the same accuracy as the CPython implementation, as they are based on the same source with the same number of iterations.

Our factorial function should be faster than the CPython implementation. :)

The `frexp` is original -- there is a Node library that implements this, but it has a silly number of dependencies that are annoying to vendor as well. It was easier to write it from scratch.


swenson:
Thanks!

PR #260, opened 2016-09-08
Basic frozenset implementation

swenson:
This has revealed some bugs in the way we handle hash table things, like dicts, sets, and frozensets. For example, we don't differentiate well between str("1") and int(1) in keys. We'll have to handle this better, by building and managing our own hash tables eventually, but this
will do for now.

I am waiting on Circle to finish running the tests so I can see all of the new "unexpected success" messages, and clean up tests. :)


swenson:
Tests pass, ready for review!

batavia/core/builtins.js
```diff
@@ -1026,7 +1035,7 @@ batavia.builtins.set = function(args,kwargs) {
     if (args && args.length > 1) {
         throw new batavia.builtins.TypeError("set expected at most 1 arguments, got " + args.length);
     }
-	return new batavia.types.Set(args[0]);
+	  return new batavia.types.Set(args[0]);
```
freakboy3742: Looks like a tab/spaces problem here - although possibly one that existed previously.

batavia/types/Set.js
```diff
@@ -7,7 +7,20 @@ batavia.types.Set = function() {
...
+                }
+            } else if (batavia.isinstance(args, [batavia.types.FrozenSet, batavia.types.List, batavia.types.Set, batavia.types.Str, batavia.types.Tuple])) {
+                var iterobj = batavia.builtins.iter([args], null);
+                var self = this;
+                batavia.iter_for_each(iterobj, function(val) {
+                    self[val] = val;
+                });
+            } else {
+                throw new batavia.builtins.TypeError("'" + batavia.type_name(args) + "' object is not iterable");
+            }
```
freakboy3742: Is there a reason this has been rolled out into the constructor, instead of the update method itself?

batavia/types/FrozenSet.js
```diff
@@ -5,11 +5,266 @@
...
+
+    FrozenSet.prototype.__or__ = function(other) {
+        throw new batavia.builtins.NotImplementedError("FrozenSet.__or__ has not been implemented");
+    };
+
+    /**************************************************
+     * Set Iterator
+     **************************************************/
+
+    FrozenSet.prototype.SetIterator = function (data) {
```
freakboy3742: Is it necessary to duplicate this implementation here? Could we borrow the one from Set (since it should be identical, AFAICT)?

freakboy3742:
Looks good for the most part; I've left a couple of notes for some areas that look like we can avoid some code duplication.

batavia/types/Set.js
swenson: Since `frozenset` is an immutable type, I thought it would be best not to have an `update` method. (`tuple` doesn't have one either.)

batavia/types/FrozenSet.js
swenson: Good call! I will move this out and dedupe.

swenson:
Rebased, deduplicated code, and addressed/answered comments and questions!

Makefile
```diff
@@ -67,6 +68,7 @@ ALL_FILES_WIN=\
     batavia\types\NotImplementedType.js \
     batavia\types\Range.js \
     batavia\types\Set.js \
+		batavia\types\SetIterator.js \
     batavia\types\Slice.js \
```
freakboy3742: This looks like more tab fun... I suspect the fix here is actually to correct all the *other* lines to actually use a tab, given make's general preference for tabs over spaces...

freakboy3742: I'm happy to do this myself - I won't block merging.

batavia/types/Set.js
freakboy3742: That makes sense for FrozenSet, but this is in Set, which *does* have an update method (and in this case, an update method that now has obvious flaws, because it doesn't handle a bunch of use cases).

swenson: Oh, yup! I can move this into the `update` method.

swenson:
Updated!

swenson:
Thanks!

PR #261, opened 2016-09-09
Implement math.log, log2, log10 for large integers

swenson:
Since we only need to calculate the logarithm to 53 bits of precision, we can just take the logarithm of the top 53 bits of precision, and add in an offset for how large the number is. Yay math!


freakboy3742:
That's a neat trick :-)

swenson:
Thanks!

PR #262, opened 2016-09-10
Implement the rest of the math module

swenson:
The gamma and lgamma implementations are adapted from public domain sources on picomath's site.

Also, we allow math module functions to accept keyword arguments by looking for `__python__`, and add an optimization to speed up the tests. (They were starting to run out of RAM in CircleCI, it seems?)


swenson:
Tests fixed, all of `math` implemented, ready for review!

swenson:
Thanks!

PR #263, opened 2016-09-12
DO NOT MERGE: Demonstrating potential way to add stdlib module

swenson:
This is more of a discussion point of a way we could load `pyc` files into batavia.

This will cause a lot of big base64 strings to be included in the `.js` file, sadly, but would reduce the need to figure out how to load files at runtime.

Note that we aren't ready to `import datetime` yet, as several key pieces prevent it from loading correctly (notably: `property` and `classmethod` are not implemented, and class inheritance doesn't quite work yet).


freakboy3742:
I see what you're doing here - the approach makes sense.

The criticism/hestiation that I have is about runtime performance. At the moment, it's just datetime, which is 76k (73k if you were drop/simplify the copyright notice). If your codebase never imports datetime, then that's 76k you didn't need to send.

76k isn't a huge problem in itself, but once the whole standard library is there, this overhead has the potential to get big, fast. 

Since the end use case doesn't involve executing live code, we're in a position to optimise the contents of stdlib.js to only contain the parts that are actually needed.

Some initial thoughts of what this might look like:

* Provide, as part of the packaging tooling, a mechanism to specify which standard library modules are required, so a custom stdlib.js can be generated.

* Allow for import-over-http. Specify an "import URL" to batavia at initialisation time; that URL will be hit if a module needs to import a system module

* Write a parser that inspects the bytecode for any import instructions, and uses that to *generate* the stdlib.js that is needed for a specific page.

These ideas aren't necessarily mutually exclusive, either. You could have a config scheme to provide a pre-packaged stdlib.js subset, falling back to an import URL if required.

Thoughts?

swenson:
I like the idea of having options -- like, `batavia-complete.js`, which has all of the standard library, or you can mix-and-match your own (I think jQuery and others do this). And I think we definitely should support import over HTTP.

If we can get donated hosting on a CDN for our stdlib.js files, even better. :)

freakboy3742:
Since this has been added in #271, I'm going to close this exploratory PR. I've opened #273 to track the import-by-http idea.

swenson:
👍

PR #268, opened 2016-09-16
Support very simple uses of super

swenson:
Still a lot to do, but this is a start.


batavia/types/Type.js
```diff
@@ -22,14 +22,19 @@ batavia.types.Object = function() {
  *************************************************************************/
 
 batavia.types.Type = function() {
-    function Type(name, bases, dict) {
+    function Type(name, bases, dict, klass) {
```
freakboy3742: This is the only thing that sticks out for me. Although you're storing it, I can't see anywhere that `__thisclass__` is used. It's a departure from the Python declaration for type() as well. Is there some reason that this is needed that I'm not seeing?

swenson: I think this was leftover from a failed attempt. I will remove it and the corresponding changes :)

swenson:
Thanks!

PR #269, opened 2016-09-20
List comparison list

jstoebel:
completed comparisons for

- List > List
- List >= List
- List < List
- List <= List

### DIFF
--- a/batavia/types/List.js
+++ b/batavia/types/List.js
@@ -71,23 +71,29 @@ batavia.types.List = function() {
     List.prototype.__lt__ = function(other) {
 
 
-        if(batavia.isinstance(other, [batavia.types.Bytes, batavia.types.Bytearray])){
+        if (batavia.isinstance(other, [batavia.types.Bytes, batavia.types.Bytearray])){
             throw new batavia.builtins.TypeError("unorderable types: list() < " + batavia.type_name(other) + "()")
         }
 
         if (other !== null) {
             if (batavia.isinstance(other, batavia.types.List)) {
-                // edge case where this=[]
-                if(this.length === 0 && other.length > 0){return true}
+                // edge case where this==[]
+                if (this.length === 0 && other.length > 0){
+                    return true;
+                }
 
-                for(var i=0; i<this.length; i++){
+                for (var i=0; i<this.length; i++){
 
                     //other ran out of items.
-                    if(other[i] === undefined){return false}
-                    if(this[i].__ne__(other[i])){return this[i].__lt__(other[i])}
+                    if (other[i] === undefined){
+                        return false;
+                    }
+                    if (this[i].__ne__(other[i])){
+                        return this[i].__lt__(other[i]);
+                    }
                 }
                 //got through loop and all values were equal. Determine by comparing length
-                return this.length < other.length
+                return this.length < other.length;
             } else {
                 throw new batavia.builtins.TypeError("unorderable types: list() < " + batavia.type_name(other) + "()");
             }
@@ -99,23 +105,29 @@ batavia.types.List = function() {
     List.prototype.__le__ = function(other) {
 
 
-        if(batavia.isinstance(other, [batavia.types.Bytes, batavia.types.Bytearray])){
+        if (batavia.isinstance(other, [batavia.types.Bytes, batavia.types.Bytearray])){
             throw new batavia.builtins.TypeError("unorderable types: list() <= " + batavia.type_name(other) + "()")
         }
 
         if (other !== null) {
             if (batavia.isinstance(other, batavia.types.List)) {
-                // edge case where this=[]
-                if(this.length === 0 && other.length > 0){return true}
+                // edge case where this==[]
+                if (this.length === 0 && other.length > 0){
+                    return true;
+                }
 
-                for(var i=0; i<this.length; i++){
+                for (var i=0; i<this.length; i++){
 
                     //other ran out of items.
-                    if(other[i] === undefined){return false}
-                    if(this[i].__ne__(other[i])){return this[i].__le__(other[i])}
+                    if (other[i] === undefined){
+                        return false;
+                    }
+                    if (this[i].__ne__(other[i])){
+                        return this[i].__le__(other[i]);
+                    }
                 }
                 //got through loop and all values were equal. Determine by comparing length
-                return this.length <= other.length
+                return this.length <= other.length;
             } else {
                 throw new batavia.builtins.TypeError("unorderable types: list() <= " + batavia.type_name(other) + "()");
             }
@@ -154,17 +166,23 @@ batavia.types.List = function() {
 
         if (other !== null) {
             if (batavia.isinstance(other, batavia.types.List)) {
-                // edge case where this=[]
-                if(this.length === 0 && other.length > 0){return false}
+                // edge case where this==[]
+                if (this.length === 0 && other.length > 0){
+                    return false;
+                }
 
                 for(var i=0; i<this.length; i++){
 
                     //other ran out of items.
-                    if(other[i] === undefined){return true}
-                    if(this[i].__ne__(other[i])){return this[i].__gt__(other[i])}
+                    if (other[i] === undefined) {
+                        return true;
+                    }
+                    if (this[i].__ne__(other[i])){
+                        return this[i].__gt__(other[i]);
+                    }
                 }
                 //got through loop and all values were equal. Determine by comparing length
-                return this.length > other.length
+                return this.length > other.length;
             } else {
                 throw new batavia.builtins.TypeError("unorderable types: list() > " + batavia.type_name(other) + "()");
             }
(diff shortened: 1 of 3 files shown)


batavia/types/List.js
```diff
@@ -78,8 +77,17 @@ batavia.types.List = function() {
 
         if (other !== null) {
             if (batavia.isinstance(other, batavia.types.List)) {
-                /* update this line to get Pythonic list < list behavior */
-                return this.valueOf() < other;
+                // edge case where this=[]
+                if(this.length === 0 && other.length > 0){return true}
```
freakboy3742: Stylistically, this sort of compression isn't desirable.
```
if (this.length === 0 && other.length > 0) {
    return true;
}
```
* space between `if` and the opening brace
* K&R style brackets, one statement per line semicolon terminated.

We don't need to save bytes or lines of code, and compressing like this impedes readability.

jstoebel: that makes sense. So just to be clear, the format you'd like to see is?

```
if (this.length === 0 && other.length > 0) {
    return true;
}
```

swenson: Yep!

jstoebel:
How are tests put together? I've not encountered tests organized like this before. From the best I can tell I might do something like this:
```
self.assertCodeExecution("""
    print('>>> x = [1,2,3]')
    print('>>> y = []')
    x = [1,2,3]
    y = []
    x < y
""")
```
But I'm not sure what else I need to do or if this is even correct. Can you enlighten me?

Thanks!

jstoebel:
I especially welcome feedback here. I'm new to this approach with tests so there might be a better way.

tests/datatypes/test_list.py
```diff
@@ -184,3 +176,88 @@ class InplaceListOperationTests(InplaceOperationTestCase, TranspileTestCase):
 
         'test_modulo_complex',
     ]
+
+class ListComparisonTests(TranspileTestCase):
```
freakboy3742: By convention, these tests are at the top of the test file, not the bottom; see [test_dict.py](https://github.com/pybee/batavia/blob/master/tests/datatypes/test_dict.py) for an example

tests/datatypes/test_list.py
```diff
@@ -184,3 +176,88 @@ class InplaceListOperationTests(InplaceOperationTestCase, TranspileTestCase):
...
+        `this` (left list) is empty.
+        """
+
+        test_str = """
+        print('>>> x = []')
+        print('>>> y = [1,2,3]')
+        x = []
+        y = [1,2,3]
+        """
+        self.assertCodeExecution(test_str+self.comparisons)
```
freakboy3742: You've got the right idea here; my only comment is about the use of the print statements. 

There are two reasons to have a print statement in the test - firstly, to validate correct behavior. This is what the "print(x > y)" is for - you're running the comparison, and making sure you get back the right answer.

The second reason is an an annotation to help debug what is going on. I'm guessing you've taken the lead from the automated tests here; they output `>>> x = ...` as the test runs. However, the key thing is *when* it does this.  The automated tests effectively output "what am I about to do", followed by "what was the result". This means that when a problem occurs, and the output doesn't match, the diff has a clear line indicating what operation was the problem. 

Based on the test you've got here, if everything works as expected, it will work fine, and keep validating that `<`/`>`/`<=`/`>=` behavior works. However, lets say some bug gets introduced and `>` stops working for some reason. Your test will output:
```
>>> x = []
>>> y = [1,2,3]
+False
-True
False
True
True
```

which isn't especially illuminating. However, if it said:
```
>>> x = []
>>> y = [1,2,3]
>>> x > y
+False
-True
>>> x >= y
False
>>> x < y
True
>>> x <= y
True
```
you can tell straight away that the problem is the > operator.

So, if you can add in some output to the template to provide these extra debug helpers, you're got it!

jstoebel:
I've added a few more tests, but I am a little unsure how to determine what else I am missing. I _think_ I have created tests to cover the following situations.


 - `this.length == 0` and `other.length > 0`
 - `this.length > 0` and `other.length == 0`
 - `this.length == 0` and `other.length == 0`
 - `this.length < other.length`
 - `this.length > other.length`
 - all items are equal
 - at least one comparable item is not equal


What am I missing? There is of course the situation when comparing differing types but this seems to result in the exception when those types are compared on their own: `[1] < ['a']` has the same exception as `1 < 'a'`

tests/datatypes/test_list.py
```diff
@@ -184,3 +316,88 @@ class InplaceListOperationTests(InplaceOperationTestCase, TranspileTestCase):
...
+#         both lists are empty
+#         """
+#
+#         test_str = """
+#         print('>>> x = []')
+#         print('>>> y = []')
+#         x = []
+#         y = []
+#         """
+#         self.assertCodeExecution(test_str+self.comparisons)
```
freakboy3742: Delete, rather than commenting out this block.

tests/datatypes/test_list.py
```diff
@@ -139,6 +139,146 @@ def test_slice(self):
...
+         print('>>> x > y')
+         x>y
+         print('>>> x >= y')
+         x >= y
+         print('>>> x < y')
+         x < y
+         print('>>> x <= y')
+         x <= y
+         """)
+
```
freakboy3742: The set of tests you have here looks good; however, you could template this for clarity. Each test follows a very repeatable pattern; ideally, that pattern would be factored out into something like:
```
    self.assertOrdering([1,2,3], [1,2,3], gt=False, gte=True, lt=False, lte=True)
```

Also - it's not enough to just do the comparison, you need to print the *result* of the comparison, because it's the output that is used as the basis for the test success/failure. so, lines 278/279 should effectively be:
```
    print('>>> x <= y')
    print(x <= y)
```

This gives you the debug output to know what the next line represents, and a printed version of that output.

jstoebel: thanks, will do. where is the proper place to implement a new assertion like `assertOrdering`? A method in `TranspileTestCase`? Some other existing class? A new class? If a new class, what file should it go in?

Regarding printing the result, you mentioned that before and I clearly didn't get it. Sorry about that.

freakboy3742: `assertOrdering` is sufficiently specific that it's OK if it's just in the TestCase in the test_list itself. Unless you can see it being used as a general capability elsewhere, it's not worth cluttering up TranspileTestCase.

batavia/types/List.js
```diff
@@ -72,14 +71,29 @@ batavia.types.List = function() {
...
+                    //other ran out of items.
+                    if (other[i] === undefined){
+                        return false;
+                    }
+                    if (this[i].__ne__(other[i])){
+                        return this[i].__lt__(other[i]);
+                    }
+                }
+                //got through loop and all values were equal. Determine by comparing length
+                return this.length < other.length;
```
phildini: It seems like we can exit faster to do the length check first? Is there some other optimization we're gaining by doing the item comparison first, and if so can we doc that?

batavia/types/List.js
```diff
@@ -91,14 +105,29 @@ batavia.types.List = function() {
...
+                    //other ran out of items.
+                    if (other[i] === undefined){
+                        return false;
+                    }
+                    if (this[i].__ne__(other[i])){
+                        return this[i].__le__(other[i]);
+                    }
+                }
+                //got through loop and all values were equal. Determine by comparing length
+                return this.length <= other.length;
```
phildini: See above

batavia/types/List.js
```diff
@@ -137,8 +166,23 @@ batavia.types.List = function() {
...
+                    //other ran out of items.
+                    if (other[i] === undefined) {
+                        return true;
+                    }
+                    if (this[i].__ne__(other[i])){
+                        return this[i].__gt__(other[i]);
+                    }
+                }
+                //got through loop and all values were equal. Determine by comparing length
+                return this.length > other.length;
```
phildini: See above

batavia/types/List.js
```diff
@@ -149,14 +193,29 @@ batavia.types.List = function() {
...
+                    //other ran out of items.
+                    if (other[i] === undefined){
+                        return true;
+                    }
+                    if (this[i].__ne__(other[i])){
+                        return this[i].__ge__(other[i]);
+                    }
+                }
+                //got through loop and all values were equal. Determine by comparing length
+                return this.length >= other.length;
```
phildini: See above

tests/datatypes/test_list.py
```diff
@@ -139,6 +139,146 @@ def test_slice(self):
             """)
 
 
+    # def test_this_empty(self):
+
+
+    def test_list_list_comparisons(self):
+
+
```
phildini: super nitpick, but this a strange amount of whitespace?

batavia/types/List.js
jstoebel: Thanks for the input! I'm a little confused about comparing lengths first. My understanding of the behavior was that we compare the first item of both lists, then the second and so on. When a non equal elements are found, their result is the result of the comparison. Comparing just the length of two lists doesn't alone tell us how they compare, we need to compare elements one by one. Example:

```
>>> [3,4] > [1,2,3]
True
```
What am I missing?

source: https://docs.python.org/3/tutorial/datastructures.html#comparing-sequences-and-other-types

jstoebel: Thanks for the input! I'm a little confused about comparing lengths first. My understanding of the behavior was that we compare the first item of both lists, then the second and so on. When a non equal elements are found, their result is the result of the comparison. Comparing just the length of two lists doesn't alone tell us how they compare, we need to compare elements one by one. Example:

```
>>> [3,4] > [1,2,3]
True
```
What am I missing?

source: https://docs.python.org/3/tutorial/datastructures.html#comparing-sequences-and-other-types

phildini: Oh! I was totally wrong, then. I thought two sequences had to be the same size to be compared. TIL!

tests/datatypes/test_list.py
```diff
@@ -4,6 +4,18 @@
...
+    def assertOrdering(self, col1, col2):
+        """
+        runs assertCodeExecution with col1 [<, <=, >, >=] col2
+        :param col1: array like
+        :param col2: array like
+        """
+        set_up = "print('>>> x = {col1}')\nprint('>>> y = {col2}')\n".format(col1=col1, col2=col2)
+        operators = ['>', '>=', '<', '<=']
+        comparisons = ["\nprint('>>> x {o} y')\nprint({col1} {o} {col2})\n".format(col1=col1, o=o, col2=col2) for o in operators]
```
freakboy3742: The only comment here is about formatting. For readability, long list comprehensions can be broken into multiple lines. You can also use the `adjust()` method in the test utilities to make the code snippet easier to read:
```
    comparisons = [
         adjust("
             print('>>> x {o} y')
             print({col1} {o} {col2})
        ").format(col1=col1, o=o, col2=col2) 
        for o in ['>', '>=', '<', '<=']
    ]
```

`adjust()` strips all left-hand-side whitespace, so you can have code snippets in multiline comments, rather than using \n in single line strings.

freakboy3742: Same comment goes for `set_up`.

jstoebel: thanks. I've made a new commit.

freakboy3742:
Fantastic! Great work @jstoebel!

PR #270, opened 2016-09-21
import this requires dict.get, str.join to be implemented

swenson:
I was working on getting the `this` module to work, but noticed these missing methods.

(There is a scoping bug that prevents the `this` module from being imported that I am working on separately.)


swenson:
Thanks!

PR #271, opened 2016-09-22
import this!

swenson:
We can now `import this` successfully.

We can now load the three modules in https://github.com/pybee/ouroboros that have no dependencies: `colorsys`, `copyreg`, and `this`.

I sketched out a simple module system, where additional files can be loaded to include compiled `pyc` files. In the future, we should be able to extend this to fetch them from external sources as well.

I modified the build so that `batavia.min.js` remains unchanged, but there is a new `batavia-all.min.js` that is also produced, which has the standard library included (including only the above three modules for now).


Makefile
```diff
@@ -87,25 +87,50 @@ ALL_FILES_WIN=\
...
+		batavia\modules\stdlib\copyreg.js \
+		batavia\modules\stdlib\this.js
+
+
+all: batavia-all.js batavia-all.min.js batavia.js batavia.min.js
+
+.PHONY: all clean stdlib
+
+stdlib:
+	python compile_stdlib.py
```
glasnt: is it the intent to have both a make point for this, and also distribute the compiled files out of this function? (You've also committed the stdlib artefacts this step creates)

swenson: Oh, good call. There is no need to check in the artifacts if we are compiling them.

I'd rather not distribute them -- they are very dependent on what version of Python was used to compile them, and we'll want to control that carefully moving forward.

swenson: Committed a fix for this this :)

batavia/core/builtins.js
```diff
@@ -35,6 +35,23 @@ batavia.builtins.__import__ = function(args, kwargs) {
     if (module === undefined) {
         module = batavia.modules.sys.modules[args[0]];
     }
+    // Check if there is a stdlib (pyc) module.
+    if (module === undefined) {
```
glasnt: using `module === undefined` twice hear makes me read it as "if this is *still* undefined, then try this other thing. Is that accurate?

swenson: Yes, that is correct.

Makefile
freakboy3742: The Python versioning is an important point. My original intention was to release major.minor that mirrored Python versions, with a .release that was specific to batavia, and always tracked the most recent (or as recent as possible) CPython point release. So, the first release of Batavia would be 3.4.1, tracking CPython 3.4.5.

Once we reach a point of partial completeness, we might need to start actively working on the individual branches. This will be especially important for the 3.6 branch, because they've completely changed the bytecode format.

compile_stdlib.py
```diff
@@ -0,0 +1,140 @@
...
+import os.path
+import py_compile
+import sys
+import tempfile
+
+enabled_modules = [
+    'colorsys',
+    'copyreg',
+    'this',
+]
```
freakboy3742: Ideally this list would be configurable. I'm wondering if it might be desirable to use this script as the general purpose "make me a .min.js" entry point allowing end users to craft a customised batavia.min.js, with the default behavior being to build the entire system.

batavia/core/builtins.js
freakboy3742: I'm pretty sure I agree with this ordering; but sanity checking to make sure it was intentional, not accidental:

This means that a module in modules namespace will be loaded before the stdlib, which means a module written in Javascript will supersede one provided as Python bytecode. So, for example, the decimal module, which was originally written in Python, could be replaced with a native Javascript version by just dropping it into the sys.modules namespace.

swenson: Yes, that was my intention.

compile_stdlib.py
swenson: I think that is something we could easily modify it for. Is that something we want to hold this PR on?

freakboy3742: No - we can improve this incrementally over time. Might even be a good thing to put on the list for a first time contributor who wants to write Python, not Javascript.

compile_stdlib.py
```diff
@@ -0,0 +1,140 @@
...
+7. Nothing in this License Agreement shall be deemed to create any
+relationship of agency, partnership, or joint venture between PSF and
+Licensee.  This License Agreement does not grant permission to use PSF
+trademarks or trade name in a trademark sense to endorse or promote
+products or services of Licensee, or any third party.
+
+8. By copying, installing or otherwise using Python, Licensee
+agrees to be bound by the terms and conditions of this License
+Agreement.
+*/
```
freakboy3742: I thought I left a comment on this one - does this need to be included in every file, or could it just be put into the *directory* (and probably this script) avoiding the duplication?

compile_stdlib.py
swenson: Awesome. We can create an issue for that once this is merged. :)

compile_stdlib.py
swenson: I don't know the answer legally, but we could probably get away with moving it to a `LICENSE` file. I'll go ahead and do that.

swenson:
Thanks!

PR #276, opened 2016-09-23
Add the bisect, operator, and stat modules

swenson:
Also allow importing from `builtins`, and added a stub of `_operator` so that `operator` will load.


swenson:
Thanks!

PR #277, opened 2016-09-24
Stub out _weakref, add in abc and _weakrefset stdlib modules

swenson:
These are three key modules in the standard library that others depend on.

Currently, it is not possible to implement `_weakref` in a reasonable manner in JS, so we stub it out for now. In the future, we might be able to do better with WeakSet (or, less likely, WeakMap), but they aren't well-supported yet.


swenson:
Thanks! I'll separate that bit out.

PR #278, opened 2016-09-24
Implement iterator for bytes

swenson:
Contains commits in https://github.com/pybee/batavia/pull/277 as well


phildini:
Can we pull the weakref diff out of the bytes iterator diff? Both look good, but you have both open right now.

swenson:
Thanks!

PR #279, opened 2016-09-25
Implement globals(), add token stdlib module

swenson:
Also implement `dict.keys()` and `dict.values()` and pass `__name__` to stdlib modules, which is needed for the `token` module.


swenson:
Oh, thanks! I was gonna rebase this later tonight. :)
On Sep 29, 2016 19:03, Russell Keith-Magee <notifications@github.com> wrote:Merged #279.

PR #280, opened 2016-09-25
Update getting started guide to include directions for ouroboros

swenson:
and running `make`

### DIFF
--- a/docs/intro/getting-started.rst
+++ b/docs/intro/getting-started.rst
@@ -16,6 +16,12 @@ The first step is to create a project directory, and clone Batavia:
     $ cd tutorial
     $ git clone https://github.com/pybee/batavia.git
 
+Batavia now requires a copy of Ouroboros (the Python standard library, written in Python) to build, so we also need to clone that.
+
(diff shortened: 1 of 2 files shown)


docs/intro/getting-started.rst
```diff
@@ -16,6 +16,12 @@ The first step is to create a project directory, and clone Batavia:
     $ cd tutorial
     $ git clone https://github.com/pybee/batavia.git
 
+Batavia now requires a copy of Ouroboros (the Python standard library, written in Python) to build, so we also need to clone that.
```
freakboy3742: No need for the tense here - someone coming to the project for the first time won't have the context of "before".

freakboy3742: A similar instruction should also be needed for the developer instructions; that section creates a completely new checkout and environment, so a fresh ouroboros checkout will also be required.

swenson: Commit added to address these comments.

swenson:
Thanks!

PR #281, opened 2016-09-27
Added shields to README.

freakboy3742:


setup.py
```diff
@@ -22,11 +22,14 @@
     ],
     license='New BSD',
     classifiers=[
-        'Development Status :: 4 - Beta',
+        'Development Status :: 2 - Pre-Alpha',
```
glasnt: Note: The badges won't update on the main README until the setup.py changes are pushed up to pypi.

PR #283, opened 2016-09-28
Prep work for implementing better dict / set / frozenset classes

swenson:
* Move `Dict` to `JSDict` -- it will still be useful in the future, as we'll still need an object-backed dictionary so that we can have a good JS to Python bridge later
* Create a new `Dict`, which merely inherits from `JSDict`
* Make a basic `hash()` implementation that delegates to `.__hash__()` or `.toString().__hash__()`
* Implement good hash function for strings
* Since the CPython and our hash values can never be expected to match, the test harness has to be modified to ignore the actual integer values. I did this by adding support for regular expression in the test substitutions


batavia/types/Dict.js
```diff
@@ -4,305 +4,11 @@
...
     function Dict(args, kwargs) {
-        Object.call(this);
-        if (args) {
-            this.update(args);
-        }
+        batavia.types.JSDict.call(this, args, kwargs);
     }
 
-    Dict.prototype = Object.create(Object.prototype);
+    Dict.prototype = Object.create(batavia.types.JSDict.prototype);
```
freakboy3742: Can you elaborate on why this separation into JSDict is necessary? It isn't obvious from this patch.

swenson: Sure thing.

The JSDict will still be necessary moving forward. It is probably going to be faster than a roll-our-own dictionary for things like our variable dictionaries. Also, we'll want to be able to have a Dict that is backed by a JS object for when we want to interact with other JS functions that aren't part of batavia (or as part of our translation).

Or maybe you were asking why we need our own Dict implementation? That's easy enough -- Javascript object's don't differentiate between two things whose `toString()` is the same, like `"1"` and `1`, or anything else that might be printed out as `1`. Javascript object keys are always strings, but Python allows any immutable type to be a key, and doesn't care about their string representations.

So, we're going to need our own Python-compliant dictionary (and likewise, set and frozenset) at some point. This is moving in that direction.

freakboy3742: So if I'm reading this right, this patch doesn't actually *do* any of that "python compliant" stuff yet - it's just a placeholder for the fact that it's going to be needed in the not-too-distant future. Is that correct?

swenson:
Yes. 
This does implement the hash() built-in. And I want the future PRs for Dict to be cleaner, so I thought I would submit this first. :) And this gives others a chance to know what I intend and comment on the approach. 
--Christopher
On Sep 28, 2016 18:35, Russell Keith-Magee <notifications@github.com> wrote:@freakboy3742 commented on this pull request.



In batavia/types/Dict.js:
>      }
 
-    Dict.prototype = Object.create(Object.prototype);
+    Dict.prototype = Object.create(batavia.types.JSDict.prototype);

So if I'm reading this right, this patch doesn't actually do any of that "python compliant" stuff yet - it's just a placeholder for the fact that it's going to be needed in the not-too-distant future. Is that correct?

swenson:
Thanks!

PR #284, opened 2016-09-30
Implement dictionaries using our own hash table impementation

swenson:
This fixes the problem where `{1: 2, "1": 3}` would produce either `{1: 2}` or `{1: 3}`. It also gives us control over hash table performance, although it is probably slower than the native JS objects.

I also fixed a few other things while I was in here:

* There were a bunch of leaky global variables I found in for-loops (found when debugging something else)
* A spurious console.log in the frozenset unmarhsaling
* A few times `args` was used in functions without checking if it is null.

Whew!

### DIFF
--- a/batavia/VirtualMachine.js
+++ b/batavia/VirtualMachine.js
@@ -1366,7 +1366,11 @@ batavia.VirtualMachine.prototype.byte_STORE_SUBSCR = function() {
 
 batavia.VirtualMachine.prototype.byte_DELETE_SUBSCR = function() {
     var items = this.popn(2);
-    items[1].__delitem__(items[0]);
+    if (items[1].__delitem__) {
+        items[1].__delitem__(items[0]);
+    } else {
+        delete items[1][items[0]];
+    }
 };
 
 batavia.VirtualMachine.prototype.byte_BUILD_TUPLE = function(count) {
--- a/batavia/VirtualMachine.js
+++ b/batavia/VirtualMachine.js
@@ -502,7 +502,7 @@ batavia.VirtualMachine.prototype.run_method = function(tag, args, kwargs, f_loca
         var payload = this.loader(tag);
         var code = batavia.modules.marshal.load_pyc(this, payload);
 
-        var callargs = new batavia.types.Dict();
+        var callargs = new batavia.types.JSDict();
         for (var i = 0, l = args.length; i < l; i++) {
             callargs[code.co_varnames[i]] = args[i];
         }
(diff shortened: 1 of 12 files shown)


batavia/types/Dict.js
```diff
@@ -2,13 +2,523 @@
...
+            updates = batavia.builtins.iter([values.items()], null);
+        } else {
+            updates = batavia.builtins.iter([values], null);
+        }
+        var i = 0;
+        var self = this;
+        batavia.iter_for_each(updates, function(val) {
+            var pieces = new batavia.types.Tuple(val);
+            if (pieces.length != 2) {
+                throw new batavia.builtins.ValueError("dictionary update sequence element #" + i + " has length " + piece.length + "; 2 is required");
```
glasnt: s/`piece.length`/`pieces.length`/g

Ref the error in CircleCI
```
### EXCEPTION ###
ReferenceError: ReferenceError: Can't find variable: piece
```

swenson:
There were a bunch of failures. :) I will comment when it is ready for review.

swenson:
All fixed up and ready for review!

batavia/VirtualMachine.js
```diff
@@ -1357,12 +1357,16 @@ batavia.VirtualMachine.prototype.byte_DELETE_ATTR = function(name) {
...
+        items[1].__setitem__(items[2], items[0]);
+    } else {
+        items[1][items[2]] = items[0];
+    }
 };
 
 batavia.VirtualMachine.prototype.byte_DELETE_SUBSCR = function() {
     var items = this.popn(2);
-    delete items[1][items[0]];
+    items[1].__delitem__(items[0]);
```
freakboy3742: Shouldn't this have the same fallback conditions as STORE_SUBSCR?

Or, alternatively, STORE_SUBSCR should be simplified to just use the Pythonic API.

batavia/VirtualMachine.js
```diff
@@ -1417,7 +1421,11 @@ batavia.VirtualMachine.prototype.byte_STORE_MAP = function() {
         case batavia.BATAVIA_MAGIC_35a0:
         case batavia.BATAVIA_MAGIC_34:
             var items = this.popn(3);
-            items[0][items[2]] = items[1];
+            if (items[0].__setitem__) {
+                items[0].__setitem__(items[2], items[1]);
+            } else {
+                items[0][items[2]] = items[1];
+            }
```
freakboy3742: What's the use case for the "javascript" style subscripting? The end goal is to get to "pure" Python objects, so the subscript shouldn't be required, right? Or is there an internal case where subscript syntax is required?

swenson: `List` doesn't have `__setitem__` (yet), and I don't think we use `List` everywhere instead of literal arrays.

batavia/VirtualMachine.js
swenson: Agreed -- I'll make this have the same fallback.

swenson: I have made this change.

swenson:
Thanks!

PR #285, opened 2016-10-01
work in progress: begining of struct_time

jstoebel:
Here is a first pass at `time.struct_time`. I would love some feedback on it.
I already know that the following are missing:

- constructor can't accept `bytearray` as its argument, I think because `bytearray` has not yet implemented `__iter__`.
- also can't accept `dict` because `.keys` has not yet been implemented.
- I haven't implemented `__reduce__` because I wasn't sure if Batavia will support pickling.

My questions:

1. What functionality did I get wrong or miss completely?
2. What tests am I not thinking of?
3. Would it make sense for me to work on `dict.keys` and/or `bytearray.__iter__`?
4. should we implement `__reduce__`?


batavia/modules/time.js
```diff
@@ -23,3 +23,75 @@ batavia.modules.time = {
         }
     }
 };
+
+batavia.modules.time.struct_time = function (sequence) {
+    /*
+        copied from https://docs.python.org/2/library/time.html#time.struct_time
+        WHAT ATTRIBUTION DOES THIS NEED?
```
freakboy3742: No copyright attribution is required, because you're not "copying" anything. You're not re-using an implementation or algorithm. You're just following a format specification. So - it's worth pointing at the documentation that defines that specification (although it would be better to point at the Python 3.4 docs, rather than Python 2), but you don't have to acknowledge copyright in this case.

batavia/modules/time.js
```diff
@@ -23,3 +23,75 @@ batavia.modules.time = {
...
+           this.push.apply(this, sequence.slice(0,9));  // only first 9 elements accepted for __getitem__
+
+            var attrs = [ 'tm_year', 'tm_mon', 'tm_mday', 'tm_hour', 'tm_min', 'tm_sec', 'tm_wday', 'tm_yday', 'tm_isdst',
+            'tm_zone', 'tm_gmtoff']
+
+            for (var i=0; i<attrs.length; i++){
+                this[attrs[i]] = sequence[i];
+            }
+
+        // TODO: need to support these types
```
freakboy3742: You can use these types in the constructor to struct_time?

README.rst
```diff
@@ -1,9 +1,22 @@
-.. image:: http://pybee.org/project/projects/bridges/batavia/batavia-72.png
-    :target: http://pybee.org/batavia
+.. image:: http://pybee.org/project/projects/bridges/batavia/batavia.png
```
freakboy3742: What procedure did you use to merge up this patch? It's a little confusing because this PR contains a bunch of code that has been reviewed and included elsewhere; `git fetch upstream && git merge upstream/master` should have introduced a merge commit, but not all the individual commits.

tests/modules/test_time.py
```diff
@@ -34,3 +31,147 @@ def test_sleep_negative(self):
...
             print('Done.')
             """)
+
+    @unittest.expectedFailure
+    def test_struct_time_valid(self):
+        """
+        valid construction
+        """
+
+        # TODO: this won't pass until the follownig types have support __getitem__()
```
freakboy3742: In this case, we should either break this test into two parts so that we can mark the failing types, but confirm the working ones actually work; or we should just implement `__getitem__`.

tests/modules/test_time.py
```diff
@@ -34,3 +31,147 @@ def test_sleep_negative(self):
...
+        bad_types = [
+            True,
+            0j,
+            1,
+            None,
+            NotImplemented
+        ]
+
+        for t in bad_types:
+            self.assertCodeExecution(struct_time_setup(t))
```
freakboy3742: In terms of debugging - this will be easier to debug if you compile a *single* string, and run one `assertCodeExecution`, rather than iterate over a loop. In the loop case, the error will always occur on line 88, and it won't necessarily be clear *which* of the types is the problem. If it's a single call rolled out internally, you'll be able to see exactly which type is the problem.

As an added bonus - starting one VM and running lots of commands is faster than starting lots of individual VMs.

freakboy3742:
Superseded by #286

PR #287, opened 2016-10-05
Normalize dictionary keys to list ordering in tests

swenson:
I noticed that this test had an unexpected success once or twice. It was because 1/6 times it would pass, depending on how the dictionary ordering.


swenson:
Thanks!

PR #288, opened 2016-10-06
Convert set and frozenset over to using a dictionary as the backend

swenson:
We will have to use a native object for some of the VM internal sets for performance reasons, as otherwise the test time more than doubles and won't run in less than an hour.


swenson:
Thanks!

PR #289, opened 2016-10-06
Ready to Merge: Add test_noargs for stuff like int() and bytes()

candeira:


glasnt:
It looks like one of the tests this PR adds into the mix has a bit of an off wording.

```AssertionError: '>>> [57 chars]TypeError\'> : abs() expected exactly 1 argument (0 given)\n\n' != '>>> [57 chars]TypeError\'> : abs() takes exactly one argument (0 given)\n\n'
  >>> f = abs
  >>> x = "_noargs (not used)"
  >>> f()
- <class 'TypeError'> : abs() expected exactly 1 argument (0 given)
?                              ^^^^^^^         ^
+ <class 'TypeError'> : abs() takes exactly one argument (0 given)
?                             +++ ^         ^^^
  
 : Global context: Error running f()
```
From the abs() function https://github.com/pybee/batavia/blob/master/batavia/core/builtins.js#L182

If you change the wording there from the current "expected exactly" to "takes exactly", this should work. I'm not sure if this will 'just work' then, or if we're going to see every function complain about slight wording variations. We did have some fun a while back where we limited diffs to only confirm the Error Type was the same, so this might be related to that issue. 

A rebase and a wording change might 'just work' :)

candeira:
Thanks a lot. It was an error in editing out the passing tests vs the ones that were expected failures. Fixed and rebased on master.

PR #291, opened 2016-10-07
mktime

jstoebel:
Here is a first stab at `time.mktime`. I welcome input on how to make it complete. One thing I'm not sure to do is how to handle a two digit year

While Python will interpret a two digit year literally, phantomjs interprets a 2 digit year as being in the 20th century not the first. Example:
```
phantomjs> new Date(70, 01, 01, 0, 0, 0, 0, 0).getTime()
2696400000
phantomjs> new Date(1970, 01, 01, 0, 0, 0, 0, 0).getTime()
2696400000     //wat
```
If handed a 2 digit year, should the function assume its in the 20th century, throw some kind of error or do something else?

### DIFF
--- a/tests/modules/test_time.py
+++ b/tests/modules/test_time.py
@@ -180,37 +180,99 @@ class TimeTests(TranspileTestCase):
 
     # TESTS FOR MKTIME
     def test_mktime(self):
-        self.assertCodeExecution(mktime_setup())
+
+        seq = (1970,1,1,0,0,0,0,0,0)
+
+        test_str = adjust("""
+        print('>>> import time')
+        import time
+        print('>>> seq = {seq}')
+        seq = {seq}
+        print(">>> time.mktime(seq)")
+        print(time.mktime(seq))
+        #################################
+        print('>>> seq = time.struct_time({seq})')
+        seq = time.struct_time({seq})
+        print(">>> time.mktime(seq)")
+        print(time.mktime(seq))
+        """).format(seq=seq)
+
+        self.assertCodeExecution(test_str)
 
     def test_mktime_bad_input(self):
         """
-        TypeError: Tuple or struct_time argument required
+        when the argument passed is not tuple or struct_time. This error is thrown
+        expected error TypeError: Tuple or struct_time argument required
         """
 
-        bad_types=[
+        seed = (1970,1,1,0,0,0,0,0,0)
 
-
+        data = [
+            False,
+            1j,
+            {key: key for key in seed},
+            1.2,
+            frozenset(seed),
+            1,
+            list(seed),
+            range(1,2,3),
+            set(seed),
+            slice(1,2,3),
+            '123456789',
+            None,
+            NotImplemented
         ]
 
+        test_str = adjust("""
+        print('>>> import time')
+        import time
+        """)
+
+        tests = [mktime_setup(str(d)) for d in data]
+        test_str += ''.join(tests)
+
+        self.assertCodeExecution(test_str)
+
+    # def test_mktime_overflow(self):
+    #     """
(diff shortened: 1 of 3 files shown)


jstoebel:
`mktime` will throw `OverflowError` if the year is < 1900. This is a somewhat arbitrary decision given that in CPython the cutoff date is determined by the underlying c library.

jstoebel:
oops one of my tests failed. I can fix it this weekend.

jstoebel:
the problem (or at least one of the problems) has to do with the fact that I don't know how to test for `OverflowError` when the platform is unknown. How can we test for this?

tests/modules/test_time.py
```diff
@@ -177,6 +176,131 @@ def test_struct_time_repr(self):
 
         self.assertCodeExecution(adjust(test_str))
 
+    # TESTS FOR MKTIME
+    def test_mktime(self):
+        """
+        tests the happy path for the mktime constructor
+        """
+
+        seq = (1970,1,1,0,0,0,0,0,0)
```
freakboy3742: A small formatting fix: spaces after commas. But something more substantial; it would be a good idea to test something other than "Midnight Jan 1 1970". This test doesn't validate that you're actually using arguments 2 & 3, or 4-9 in the right order - you might, for example, be getting the month and the day the wrong way around (you're not, but hopefully you can see where the *potential* bug might lie).

tests/modules/test_time.py
```diff
@@ -177,6 +176,131 @@ def test_struct_time_repr(self):
...
+
+        self.assertCodeExecution(test_str)
+
+    @unittest.expectedFailure
+    def test_mktime_overflow_error(self):
+        """
+        tests OverflowError on dates earlier than an arbitrarily defined date
+        for now this is defined as 1900-01-01
+
+        currently this can't be garuneteed to pass because it doesn't not account for behavior across platforms.
```
freakboy3742: I'd suggest the best approach here is to use `assertJavaScriptExecution()` rather than `assertCodeExecution`. The rest of this test suite will validate that the behavior of mktime is consistent with CPython; all this test needs to do is ensure that an error will be raised under known conditions. It's worth a documentation note to describe *why* this test can't be performed in a cross platform way; but that's all you really need to do.

jstoebel: thanks for this feedback! I will work on it and send a new PR.

jstoebel:
I'm now using `assertJavaScriptExecution` and two tests are breaking (see below). They are passing on my mac but fail in the CI. Out of curiosity I fired up a Docker image with Ubuntu/phantomjs and compared the output of `Date`

### on my mac, 10.11.6
```
phantomjs> new Date(1900, 0, 1,0,0,0,0).getTime() / 1000
-2208970800
```
### inside Docker image, Ubuntu/PhantomJS
```
phantomjs> new Date(1900, 0, 1,0,0,0,0).getTime() / 1000
-2208988800
```
I will dig a little more to figure out what's going on but if anyone knows or has a lead I'd love to hear it.

jstoebel:
I modified `assertJavaScriptExecution` to expect either matching output or not matching output. This is how I assert that mktime won't throw an error. What are your thoughts about this? I could also predict the actual returned value but this seemed like a simpler approach.

jstoebel:
I also wrote a blog post about this PR if that's interesting to anyone. http://www.jstoebel.com/implementing-mktime-in-batavia/

glasnt:
This is looking really good @jstoebel! I'll let @freakboy3742 have the final word, though. 

I'd also love to add your blogpost to the [website resources](http://pybee.org/community/resources/) list, with your consent.

jstoebel:
@glasnt of course! Go ahead and use my blog post.

PR #292, opened 2016-10-07
Starting work on compile function

swenson:
I've started digging into compiling Python code.

This copies over the basic skeleton of CPython `compile()` (though with most functions not actually implemented).

I've also started working on translating the CPython hand-written lexer into JS. And it works on a trivial example (`x = 1`)!

This code is certainly a giant work-in-progress, and I won't be offended if we don't want to merge it in yet, as the indentation is all over the place. But I just wanted to check in with where I am going.


swenson:
I'd also want to comment that the hand-written lexer can probably be simplified a lot. I removed some functionality as I went, such as support for file input.

freakboy3742:
Ok. Um. Wow. Someone pass the smelling salts... I'm going to need a moment to compose myself...

swenson:
Oh there'll be a lot more coming. Might want to stock on smelling salts. :)

swenson:
This is a stab at https://github.com/pybee/batavia/issues/294. We probably will need to break this up into smaller pieces eventually. But for now, it makes some sense I think to work on it all together.

swenson:
Going to discontinue this now and start splitting it up into pieces.

PR #298, opened 2016-10-11
Add information about setting up Python 3.4

glasnt:
I'm not 100% happy with this as is. `pyenv` works in OSX fine, but I'm not sure how it fares in other environments, and I don't want to be making recommendations that don't work in all environments.

### DIFF
--- a/docs/internals/contributing.rst
+++ b/docs/internals/contributing.rst
@@ -17,7 +17,7 @@ instead of using the official PyBee repository, you'll be using your own
 Github fork.
 
 As with the getting started guide, these instructions will assume that you
-have Python 3.4, and have virtualenv available for use.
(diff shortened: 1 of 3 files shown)


swenson:
I think `pyenv` is fine on Linux and most other *nixes, but I don't think it works in Windows.

glasnt:
@swenson if there's a known 'best helper' for Windows, I can add that to the doc. Even if it's just a link out to the best practice. Having a self-contained tutorial is great, but I don't want to be duplicating too much copy from other more complete resources

glasnt:
I have a better idea on how to implement this. Will open another PR later

glasnt:
https://github.com/pybee/pybee.github.io/pull/49

PR #299, opened 2016-10-11
Added need for SSH key

williln:
Added a quick line at the top to specify the need for having an SSH key set up, since I did not and that was one of my problems this week :) LMK if this kind of thing is so obvious as to not need to be documented! But it was new to me.

### DIFF
--- a/docs/intro/getting-started.rst
+++ b/docs/intro/getting-started.rst
@@ -2,7 +2,7 @@ Getting Started
 ===============
 
 In this guide we will walk you through setting up your Batavia environment for
-development and testing. We will assume that you have a working Python 3.4,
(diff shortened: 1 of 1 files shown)


docs/intro/getting-started.rst
```diff
@@ -2,7 +2,7 @@ Getting Started
 ===============
 
 In this guide we will walk you through setting up your Batavia environment for
-development and testing. We will assume that you have a working Python 3.4,
+development and testing. We will assume that you have an `SSH key <https://help.github.com/articles/generating-an-ssh-key/>`_ set up for your GitHub account, have a working Python 3.4,
```
glasnt: So, I know we hit this error when we were trying to debug your issues, @williln, but a github SSH configuration isn't required by any instruction on this particular page (all `git clone` references use HTTPS). There are two references in `contributing.rst` that I need to clean up, but I'm going to reject this PR based on the above.

PR #307, opened 2016-10-17
Added the base64 module

martini97:
Added the base64 module with the b64encode and b64decode.
Signed-off-by: Alessandro Fernandes Martini <alessandrofmartini@gmail.com>

### DIFF
--- a/batavia/modules/base64.js
+++ b/batavia/modules/base64.js
@@ -1,0 +1,64 @@
+batavia.modules.base64 = {
+	__doc__: "",
+	__name__: "base64",
+	__file__: "base64.js",
+	__package__: "",
+
+	b64encode: function(data, altchars){
+		altchars = altchars || 0;
+		if (altchars === 0){
+			return new batavia.types.Str(window.btoa(data));
+		}
+	},
+
+	b64decode: function(data){
+		if (data.len % 4 !== 0) {
+			throw new batavia.builtins.ValueError("Incorrect padding");
+		}
+		return new batavia.types.Str(window.atob(data));
+	},
+
+	_a85chars: function(){},
+	_a85chars2: function(){},
+	_b32alphabet: function(){},
+	_b32rev: function(){},
+	_b32tab2: function(){},
+	_b85alphabet: function(){},
+	_b85chars: function(){},
+	_b85chars2: function(){},
+	_b85dec: function(){},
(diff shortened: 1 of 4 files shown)


swenson:
Looks like the base64 tests for this are currently failing:

```
FAIL: test_b64decode (tests.modules.test_base64.Base64Tests)
  File "/home/ubuntu/batavia/tests/modules/test_base64.py", line 20, in test_b64decode
    """)
  File "/home/ubuntu/batavia/tests/utils.py", line 504, in assertCodeExecution
AssertionError: "### EXCEPTION ###\nTypeError: TypeError: [64 chars]:3\n" != "b'Foo'\n"
+ b'Foo'
- ### EXCEPTION ###
- TypeError: TypeError: undefined is not an object (evaluating 'obj[attr]')
-     test.py:3
 : Global context

FAIL: test_b64encode (tests.modules.test_base64.Base64Tests)
  File "/home/ubuntu/batavia/tests/modules/test_base64.py", line 13, in test_b64encode
    """)
  File "/home/ubuntu/batavia/tests/utils.py", line 504, in assertCodeExecution
AssertionError: "### EXCEPTION ###\nTypeError: TypeError: [64 chars]:3\n" != "b'Rm9vOQ=='\n"
+ b'Rm9vOQ=='
- ### EXCEPTION ###
- TypeError: TypeError: undefined is not an object (evaluating 'obj[attr]')
-     test.py:3
 : Global context
```

batavia/modules/base64.js
```diff
@@ -0,0 +1,64 @@
...
+	standard_b64decode: function(){},
+	standard_b64encode: function(){},
+	struct: function(){},
+	test: function(){},
+	urlsafe_b64decode: function(){},
+	urlsafe_b64encode: function(){},
+
+};
+
+batavia.modules.base64.isclose.__python__ = true;
```
swenson: If I had to guess, I'd think this is the problem? There is no `isclose` function (like there is in the `math` module), and we don't need to support keyword arguments I would guess, so no need to mark `__python__` to true in any of the modules.

martini97: Yes, it was, but now I'm having some trouble with conversion, since the functions should return bytes, and the atob and btoa methods return strings I have to convert them, and this is throwing some mistakes regarding the type

PR #314, opened 2016-10-18
Implementing int to complex addition

barmvicente:


swenson:
Is this the same as #321?

PR #315, opened 2016-10-18
Implement bool ge str, bytes, bytearray, class.

ramiroluz:


glasnt:
I've just kicked off another run of the build for this. Once it's green, 👍

PR #316, opened 2016-10-18
Added the urlsafe_b64encode method to base64

martini97:
Signed-off-by: Alessandro Fernandes Martini <alessandrofmartini@gmail.com>

### DIFF
--- a/batavia/core/builtins.js
+++ b/batavia/core/builtins.js
@@ -732,8 +732,9 @@ batavia.builtins.help.__doc__ = 'In Python, this is a wrapper around pydoc.help.
 batavia.builtins.hex = function(args) {
     if (args.length !== 1) {
         throw new batavia.builtins.TypeError("hex() takes exactly one argument (" + args.length + " given)");
-    }
-    return "0x" + args[0].toString(16);
+    };
+    var int = args[0].val.c[0]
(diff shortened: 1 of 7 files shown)


batavia/core/builtins.js
```diff
@@ -732,8 +732,9 @@ batavia.builtins.help.__doc__ = 'In Python, this is a wrapper around pydoc.help.
 batavia.builtins.hex = function(args) {
     if (args.length !== 1) {
         throw new batavia.builtins.TypeError("hex() takes exactly one argument (" + args.length + " given)");
-    }
-    return "0x" + args[0].toString(16);
+    };
+    var int = args[0].val.c[0]
```
swenson: This doesn't print the entire integer -- it will only work for very small numbers.

Is this tested in the builtins tests?

For example,

```
>>> print(hex(0xb2594c0a7500649c618b365c0663d9d8556be9784939edb663267c0a8288))
```

Should print "0xb2594c0a7500649c618b365c0663d9d8556be9784939edb663267c0a8288", but this prints only "0x7b".

freakboy3742: Patch incoming; no, it wasn't tested. It is now, with 0xf, 0xdeadbeef, and a really big int.

Issue #275, opened 2016-09-23
`compile_stdlib.py` improvements


swenson:
The `compile_stdlib.py` is a program that converts standard library `.py` files into `.js` files by compiling them to bytecode (`pyc`), and then encoding that in JS.

There are a number of improvements that we would like to make to this tool.

* [ ] Accept arguments to specify what libraries to include
* [ ] Modify it to produce the `batavia.js` file itself, patching it together, rather than doing it in the `Makefile`
* [ ] Have arguments for common library collections (e.g., `--basic-stdlib` for a lighter standard library, once we have more of it implemented)
* [ ] Your other cool ideas go here

freakboy3742:
Another idea: make the path to the Ouroboros directory a configuration flag.

freakboy3742:
And another: a configuration file storing the build options that can be passed in at runtime. This enables the build options to be version controlled as part of a build/deployment system.

parvmor:
Hey! I am a first timer and want to contribute towards this project. So, could you explain a little bit more about this issue?

freakboy3742:
Hi @parvmor - Batavia consists of two parts:
1) the core language interpreter
2) the Python standard library.

Much of the Python standard library doesn't have to be re-written - we can just use the code from CPython itself. This is because much of CPython's standard library is written in Python. The pure Python implementation of the CPython standard library is contained in the Ouroboros project.

However, in order for Batavia to *use* that code, we need to process the contents of Ouroboros into bytecode format so it can be packaged with Batavia. This is done by the `compile_stdlib.py` script. 

So - the initial task is to modify `compile_stdlib.py` to include more packages from the standard library provided by Ouroboros.

The secondary task is to provide a way to customise `compile_stdlib.py` so that an end user can customise which parts of the standard library are included. If you're delivering a web app, size matters, so there's no point delivering parts of the standard library that aren't going to be used. Being able to specify or customise the parts of the standard library that will be included in your Batavia build would be helpful.

Does that clarify the requirements enough for you to get started?

mrPain89:
Hi @freakboy3742 , I'd like to contribute on this issue. My question is : would this be helpful to add an argument to compile_stdlib that lists available modules so the end user doesn't have to crawl through the paths to see what's in ouroboros module folder. 

2nd question is : In ouroboros module folder there is also folders (like `urllib`), we should be able to include libs that are in these subfolders. 
[...]

PR #317, opened 2016-10-18
Parse args with click

tonnydourado:
This implements the argument parsing logic with click. refs #275

### DIFF
--- a/compile_stdlib.py
+++ b/compile_stdlib.py
@@ -15,45 +15,23 @@ import py_compile
 import sys
 import tempfile
 
-
-def parse_args():
-    parser = argparse.ArgumentParser(description=__doc__)
-    parser.add_argument('modules', metavar='module', nargs='*',
-                        help='source modules to compile')
-    parser.add_argument('--source', help='location of the ouroboros source files')
-
-    args = parser.parse_args()
-
-    enabled_modules = args.modules or [
-        '_weakrefset',
-        'abc',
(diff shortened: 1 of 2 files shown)


tonnydourado:
The main advantage of click is that it allows you to express the validations for the command line arguments in a declarative way, which tends to be shorter and easier to read. With argparse all the checks like 'is this a valid path?', 'does this path exists?', 'is it a directory', have to be manually coded. On click, it's just a parameter in a decorator.

It also supports sub commands, which allows you to build pretty complex CLIs with just a bunch of functions and some decorators.

That said, this script has just two parameters as it is, so, might not be worth yet. But as more of the ideas from #275 are implemented, the advantages might become more worth adding the dependency. 

On the other hand, I tried to implement that idea of having the arguments come from a config file and it didn't play out as smoothly with click as I hoped, so, there's that too.

freakboy3742:
@tonnydourado Thanks for that clarification.

@glasnt @swenson @phildini Do you have any thoughts on this?

swenson:
I quite like `click`'s style. However, I do agree that it seems premature to use it, as we don't have a complex CLI or enough options to justify the extra dependency.

glasnt:
Thank you for your time and explanation, @tonnydourado. Based on the comments above, I'm going to close out this PR. 

If we end up with a more complex CLI, we can revisit this.

PR #338, opened 2016-10-26
Added some new functions and fixed the old ones

martini97:
Signed-off-by: martini97 alessandrofmartini@gmail.com

### DIFF
--- a/batavia/modules/base64.js
+++ b/batavia/modules/base64.js
@@ -4,9 +4,20 @@ batavia.modules.base64 = {
 	__file__: "base64.js",
 	__package__: "",
 
-	b64encode: function(data){
-		var data_str = String.fromCharCode.apply(null, data.val)
+	b64encode: function(data, altchars){
+		altchars = altchars || 0;
+		if (altchars !== 0) {
+			altchars = String.fromCharCode.apply(null, altchars.val)
+		};
+		var data_str = String.fromCharCode.apply(null, data.val);
 		var encode = window.btoa(data_str);
+		if (altchars === 0) {
+		} else if (altchars.length === 2) {
+			encode = encode.replace(/[+]/, altchars[0]);
+			encode = encode.replace(/[/]/, altchars[1]);
+		} else {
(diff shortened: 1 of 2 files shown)


phildini:
Hi there. Thanks for the PR! At first glance, I can't really tell why this version is improved over the old version. If you could add some tests that now pass with this version (that preferably wouldn't have passed with the old version) that would be great!

At the very least, an explanatory comment or two feel necessary.

martini97:
The previous version didn't have the altchars for b64encode and b64decode, no validate for b64decode, and the urlsafe encode and decode wasn't replacing the + and /. I will insert the needed tests.

freakboy3742 (changes requested):
I must be missing something; I can see the extra `altchars` argument, but the extra test cases you provide all pass without making any changes to the base64 methods (at least, they do in my testing)

I'm not denying that the changes are needed; but the tests you provide need to protect against regressions by demonstrating failing behavior before you make them pass.

martini97:
@freakboy3742 I'm trying to add the tests, but there is some ASCII problem with the ascii, I've tried running it in python and it works so I'm a little lost,if you would like to see the errors they are here: http://pastebin.com/WtcdE11F

phildini:
Hey there! So sorry we didn't see this sooner. If you're still interested in moving forward with this PR, would you mind rebasing and fixing the conflicts?

shaib:
I've squashed and fixed the tests into one commit, based on current master, at https://github.com/shaib/batavia/tree/martini97-b64. The addition causes one test to fail.

The current implementation of b64encode and b64decode is completely different than what it was at the time this PR was written; they now rely on the base64-js library, which has no room for `altchars`. So a design decision needs to be made, if a written-from-scratch implementation is wanted. It seems one is required for 100% compatibility with CPython. If we do, we should probably completely replace the current implementation with something like we see here. We would also need to add tests for `altchars` on `b64decode()`.

phildini:

PR #339, opened 2016-10-28
localtime

jstoebel:
This is a draft of `time.localtime` Feedback welcome!


jstoebel:
I'm still lost as to the solution to fixing this. Does anyone have any advice on how to make this test pass?

swenson:
Oh, I see this error now:

```
Your build has exceeded the memory limit of 4G on 1 container
```

Huh. I'll take a deeper look later today.

swenson:
Looks like since this is on your own fork, you'll have to redo the tests: just click "Rebuild" on this page: https://circleci.com/gh/jstoebel/batavia/49?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

jstoebel:
its rebuilding now. there should be an answer either way shortly

swenson:
I'm guessing that the inclusion of `moment` and `moment-tz` makes the JS process too large, since we spawn 11,000+ of them.

Would it be possible to lazy-load the `moment` libraries, and only load them up if the `time` module itself is imported?

jstoebel:
ok I've remove `moment` from vendoring and now require it inside `time`. Was this the right thing to do?

jstoebel:
thanks for this feedback. I'm having a hard time understanding how to do this. I've tried.

```
require('moment')
require('batavia/vendor/node_modules/moment')
require('batavia/vendor/node_modules/moment/moment.js')
```
How do I do this?

swenson:
I don't think it will be as straightforward as that, unfortunately. The `require` function may only work in the `vendor` directory -- I'm not sure.

I was thinking more along the lines of:

* Convert the `require('moment')` vendor line to something more like, `function() { if (moment == null) { moment = require('moment'); } return moment; }`, i.e., so that it isn't initialized, but is instead lazy initialized when first used.
* When the `time` module is imported, it initializes the `moment` library by calling that function.

I'm not 100% sure how well this will work though.

jstoebel:
@swenson thanks I'll give it a crack

jstoebel:
@swenson looks like its still timing out. I'm stumped. Any other ideas?

swenson:
I suspect that this the `moment` and `moment-tz` libraries are just causing us to use too much memory with so many tests.

We could try switching on the garbage collector in the `gc` module, and running `gc.collect()` after every suite, perhaps?

Now that I think about it, CPython doesn't ship with all of the `tz` data preloaded -- that's in a separate package. Though I don't know how they deal with finding out the DST information...

jstoebel:
thanks, I will take a look at this and see if I can find a solution.

jstoebel:
@swenson when you said
> We could try switching on the garbage collector in the gc module, and running gc.collect() after every suite, perhaps?

did you perhaps mean switching _off_ the garbage collector? I thought the gc runs automatically by default.

swenson:
@jstoebel I noticed that the out of memory condition was triggered once on `master` as well, so it may be tricky to fix here.

I thought that garbage collection was optional in CPython -- it normally uses reference counting. So, I was hoping switching it on might do better, or at least differently.

swenson:
I think you used the `gc` module correctly, so I'd say that this didn't help.

swenson:
I think the most logical thing to do at this point would be to put the `time` module into the `EXTRA_FILES` target in the `Makefile`, and, by default, run tests with only the base `batavia.min.js`.

For the `time` and other standard library tests, we'd want to override the behavior to load `batavia-all.min.js`, but just for those tests.

Hopefully that will reduce the memory usage enough for the tests to run.

jstoebel:
Thanks @swenson. How exactly would I make this change? In `Makefile`  I removed `time.js` from `BASE_FILES` and added it to `EXTRA_FILES` but how do I  "verride the behavior to load batavia-all.min.js, but just for those tests."

swenson:
@jstoebel Sorry for being quiet. Have been busy.

Try to rebase this against `master` now, and see if it passes. I turned off a bunch of services that were running by default and consuming a bunch of memory.

jstoebel:
looks like I messed something up in this last rebase. I have to go to work right now, but I will fix tonight or tomorrow.

freakboy3742:
Closing in preference to #377.

PR #344, opened 2016-11-03
Fixes for None and getattr handling.

freakboy3742:


batavia/core/builtins.js
```diff
@@ -1415,5 +1425,5 @@ for (var fn in batavia.builtins) {
     batavia.builtins[fn].__python__ = true;
 }
 
-batavia.builtins.None = null;
+batavia.builtins.None = batavia.types.NoneType;
```
swenson: This seems odd to me. In CPython, `None` is an instance of `NoneType`, but not equal to it.

Also, I think when making this, we'll have a lot of extra cleanup to do. (For example, searching for `== null` in the codebase has dozens of hits that appear to be `None` checks.

freakboy3742: There are a bunch of other cleanups - I've just pushed another update to fix some more problems that I found.

The problem we have here is that we can either put `value === null` checking in every comparison check, or we can make `None` an object that behaves like every other object. I've hit this in preparing toga-django; the fact that `None` isn't an object causes all sorts of edge cases that are hard to find and diagnose.

I agree that `None = batavia.types.Nonetype` is odd - it's like that because NoneType isn't a Type as currently implemented. I can fix that, though.

PR #345, opened 2016-11-04
Compile: lexical analysis

swenson:
This is mostly identical to the CPython implementation of the Python lexer, reimplemented in JS.


swenson:
Whew, tests pass, ready for review!

batavia/modules/_compile/tokenizer.js
```diff
@@ -0,0 +1,1095 @@
...
+        nonascii = 1;
+    }
+    c = tok.tok_nextc();
+}
+tok.tok_backup(c);
+if (nonascii && !verify_identifier(tok)) {
+    return [ERRORTOKEN, p_start, p_end, 10];
+}
+
+  // async/await parsing block.
```
glasnt: Is this giant block of commented-out code intended to be left in?

batavia/modules/_compile/tokenizer.js
```diff
@@ -0,0 +1,1095 @@
...
+
+Tokenizer.prototype.tok_nextc = function() {
+var tok = this;
+
+if (tok.cur != tok.inp) {
+    return tok.buf[tok.cur++]; /* Fast path */
+}
+return EOF;
+
+// for (;;) {
```
glasnt: Same again here. I know this is compile parsing code, so giant blocks of comments may well be somewhat intended, but it doesn't appear to be useful in this case. Happy to be corrected if wrong

batavia/modules/_compile/tokenizer.js
swenson: It is. We don't support `async` yet, but we will, and then we'll need this block of code to handle it.

batavia/modules/_compile/tokenizer.js
swenson: You're right. I don't think we need this block of code anymore. It is intended if we are reading the file chunks at a time, like from a disk, but we assume we have it as a string already. I'll delete it.

swenson:
@glasnt We're not *quite* there yet. This is just the tokenizer, that splits a Python program up into tokens. The next few PRs will parse the tokens into an AST, and then generate the `.pyc` bytecode. When we get to those points, I totally agree we should add thorough functional tests. For now, I don't think we can. Yet! :)

glasnt:
@swenson ah, righto then!
Would you like a merge once the latest CI test goes green? :)

swenson:
@glasnt Sure!

swenson:
Thanks!

Issue #294, opened 2016-10-08
Compile Python into bytecode in batavia directly


swenson:
In order to be self-hosted (and run without CPython also being present), we need to be able to compile Python into bytecode ourselves.

There are four big pieces to making this happen:

1. Lexical analyzer (convert source to series of simple tokens).
2. Write code generator for grammar to autogenerate a lot of JS for us.
3. Implement the parser to convert tokens into an AST (with the help of the grammar).
4. Convert the AST into Python byte code (generate).

In CPython, all of this is done in C.

alexcondrea:
Anyone still working on this? Would like to contribute.

PR #349, opened 2016-11-05
[WIP] Make Batavia installable via NPM

idan:
- [x] Create a valid package.json
- [x] remove `vendor` dir and add those things as proper dependencies
- [ ] does it `npm install` cleanly?


candeira:
+1

I'm the author of the Batavia vendoring band-aid. All I know about JS packaging is what I learnt while implementing vendoring. However, this is what I would have done instead of vendoring if I had known how. 

Batavia is a JS project (that happens to implement Python), and it should live in JS-land.

In particular, once @swenson 's https://github.com/pybee/batavia/issues/294 is implemented, people should eventually be able to do one of two things:

1. Write "text/javascript" scripts and ship them to the browser to be compiled there and interpreted by the packaged batavia.min.js, or
2.  have their Python code compiled by Batavia at build time, using something like the Google Closure compiler to perform dead code elimination, and shipping to the browser only the bits of the Python standard library that are used by the script.

Point 1) above is enabled by #294, but the door to 2) is opened by the present PR.

freakboy3742:
Ok - it appears that we have now hit the limit of what PhantomJS can do without causing memory blowouts.

The good news is that one of the side effects of this patch is that we can abandon PhantomJS and just use node directly! Instead of creating fake browser content and inserting batavia.js, we *should* be able to just `require('batavia.js')` and invoke the test suite.

However, I'm still coming up to speed with node and webpack, so I need some assistance.

If you run `webpack` to build batavia.js, you should then be able to write a `sample.js` file like the following:
```
var batavia = require('batavia.js');

var modules = {
    'sample': '7gwNCsJuLFgWAAAA4wAAAAAAAAAAAAAAAAIAAABAAAAAcw4AAAB' +
        'AAcgIAAADAQABZAEAUykCegtoZWxsbyB3b3JsZE4pAdoFcHJpbnSpAHICAA' +
        'AAcgIAAAD6PC92YXIvZm9sZGVycy85cC9uenY0MGxfOTc0ZGRocDFoZnJjY' +
        '2JwdzgwMDAwZ24vVC90bXBmbjQwZHJuc9oIPG1vZHVsZT4BAAAAcwAAAAA='
};

var vm = new batavia.VirtualMachine(function(name) {
    return modules[name];
});

vm.run('sample', []);
```

and it should print "Hello world".

However, based on the current webpack definition, this results in `batavia.VirtualMachine` being undefined; If I change the `libraryTarget` to be `umd`, `batavia.VirtualMachine` is defined, but an error is raised inside the dis module because `batavia` isn't defined at all.

Can anyone with node/webpack-fu help out here?

alonisser:
@freakboy3742 I have some webpack experience and can take a look.  Can you please clarify:
* Should I check this specific PR or was the webpack problem before the PR?
* Is this process supposed to replace the make step while building the project? 
* What is phantomjs actually used for here or can I just ignore that part?

Thanks

freakboy3742:
@alonisser Thanks for taking a look! Regarding your questions:

1) This PR is the only place webpack is being used as present. 

2) Yes - we were using make previously; this PR is trying to replace the make script with a webpack script. This will (hopefully) make it easier for others to include batavia in their projects.

3) PhantomJS is being used to run the code, because the way we were packaging the code previously required a browser for the code to run. That's one of the things I'm hoping to fix with this path - PhantomJS is a bit of a leaky beast, but if by packaging differently, we can "require('batavia)", that would allow us to move to node to run batavia test code, which would be vastly preferable.

If you've got any other questions, I'm happy to answer them; either here, or in our [gitter room](https://gitter.im/pybee/general)

PR #351, opened 2016-11-05
Add error display to testbed

adamchainz:
Makes it easier to try out Batavia by showing failures to compile.

### DIFF
--- a/testserver/requirements.txt
+++ b/testserver/requirements.txt
@@ -1,1 +1,1 @@
-django==1.8
+django==1.10
--- a/testserver/testbed.html
+++ b/testserver/testbed.html
@@ -244,7 +244,7 @@ DAEDABcBDQEdAgYB
 {{ othercode }}
     </script>
 
-    {% if customcode %}
+    {% if customcode and customcode.compiled %}
     <script id="batavia-customcode" type="application/python-bytecode">
 {{ customcode.compiled }}
     </script>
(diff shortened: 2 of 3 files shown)


adamchainz:
screen shot 2016-11-05 at 14 44 21

glasnt:
Unsure if @freakboy3742 has reviewed this IRL, but here's my take. 

I don't think we rely on anything in Django 1.8, so the bump to 1.10 seems fine to me. 

Can we do something... "fancy" with the error message to swap out the "/var/folders" file location? I know we're doing some funky stuff behind the scenes anyway, but removing that local file name that presumably doesn't exist long after execution might be an idea to make the interface feel nicer.

adamchainz:
I think something can be done yes with just a `str.replace`, but it might be a bit flakey.

glasnt:
@adamchainz Yeah, I think it's best just to leave it be. 

@freakboy3742 your final say on this would be appreciated.

freakboy3742:
I love it! In particular I like the move of the output to be *above* the code window - from a UX perspective, I find that better than the "buried" location output of stdout.

If you're looking for next steps with this, might I suggest:
* Redirecting stderr as well as stdout, so that when the code throws an error at runtime, it appears in the same error block (This might involve reworking the run method to capture exceptions)
* Tweaking the CSS so that stdout is in a float; it would appear *above* the code block on a narrow screen, but on a wide screen, the display goes split screen with code on the left, output on the right.

PR #370, opened 2016-11-09
Turn off unnecessary memory hog services in CircleCI config

swenson:
These services are unnecessary for our test setup, and they consume more memory than our Python tests by quite a margin.

This should help with some of the other PRs that increase memory usage of our test scripts.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,6 +9,11 @@ dependencies:
     - test -e ouroboros || git clone https://github.com/pybee/ouroboros.git
 
 test:
+  pre:
+    - sudo service couchdb stop || true
+    - sudo service mongod stop || true
+    - sudo service mysql stop || true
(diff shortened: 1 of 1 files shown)


swenson:
It's a bad idea because it will break all the time, and then our jobs would be keeping the CI servers up instead of writing and reviewing code. :)

Thanks!

glasnt:
I agree with @swenson. Trust me, anything we can offload to a third-party is more time we can spend on developing code. Even if we have to fine-tune things occasionally.

freakboy3742:
I know what you're saying. We're just hitting so many annoying limits with both Circle and Travis (worker count limits, CPU limits, job duration limits) and we still can't deploy Beefore... and all we need is "start a container with X configuration, check out Y, and run Z".

PR #373, opened 2016-11-13
Correct relative imports.

freakboy3742:
This also fixes a few edge cases for importing native packages and deep stdlib packages.


batavia/VirtualMachine.js
```diff
@@ -9,7 +9,21 @@ batavia.VirtualMachine = function(loader) {
 
     if (loader === undefined) {
         this.loader = function(name) {
-            return document.getElementById('batavia-' + name).text.replace(/(\r\n|\n|\r)/gm, "").trim();
+            var element = document.getElementById('batavia-' + name);
```
phildini: This code looks good from a code clarity perspective, not quite sure what the code is doing.

batavia/core/builtins.js
```diff
@@ -24,158 +24,198 @@ batavia.builtins.<fn>.__doc__ = 'docstring from Python 3.4 goes here, for docume
 */
 
 batavia.builtins.__import__ = function(args, kwargs) {
+    // console.log("IMPORT", args[0], args[1], args[4]);
```
phildini: This makes me think we should have some sort of "DEBUG" mode in batavia that we can turn on and off through init flags or args. 

This isn't me complaining about the comment, to be clear, this is me wishing I could conditionally enable the logging.

batavia/VirtualMachine.js
freakboy3742: It's finding the `<script>` block that contains base64 encoded bytecode, getting the text contents of that block, and stripping out all the whitespace. That content is the payload that can then be decoded from base64 and passed to the marshaller.

I'll add this as a comment in the code for posterity.

batavia/core/builtins.js
freakboy3742: Agreed. Any ideas on the best way to handle this?

PR #374, opened 2016-11-13
Windows readme updates

clach04:
venv setup is slightly different under Microsoft Windows, along with `make` call.

Port for test server updated to reflect the actual port used.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -54,16 +54,27 @@ Downloading and Installing
    $ git clone https://github.com/pybee/batavia
    $ git clone https://github.com/pybee/ouroboros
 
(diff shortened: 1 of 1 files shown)


clach04:
NOTE following updated readme works for me with Miniconda3 (Python 3.5.1), however the virtualenv setup as documented didn't work for me, I had to issue:

    C:\Miniconda3\python.exe C:\Miniconda3\virtualenv.py --system-site-packages venv

To avoid:

    ImportError: No module named 'pip'

pip really is available (and `--system-site-packages venv` shows that works) so I'm slightly confused over this, this _may_ be limited to my machine, hence not including in the change.

PR #380, opened 2016-11-27
Escape regex string

jpietrangelo:
Small edits to testserver/testbed.html and batavia/VirtualMachine.js. HTML file needed an opening <html> tag. JS file needed the regex code to include backslashes to escape the existing backslashes.

### DIFF
--- a/batavia/VirtualMachine.js
+++ b/batavia/VirtualMachine.js
@@ -9,7 +9,7 @@ batavia.VirtualMachine = function(loader) {
 
     if (loader === undefined) {
         this.loader = function(name) {
-            return document.getElementById('batavia-' + name).text.replace(/(\r\n|\n|\r)/gm, "").trim();
+            return document.getElementById('batavia-' + name).text.replace(/(\\r\\n|\\n|\\r)/gm, "").trim();
         };
     } else {
         this.loader = loader;
--- a/testserver/testbed.html
+++ b/testserver/testbed.html
@@ -1,4 +1,5 @@
 <!doctype html>
+<html lang="en">
 <head>
     <title>Batavia testbed</title>
 


glasnt:
Hi @jpietrangelo! Since we don't have any automated testing on the testbed, could you provide a before and after of what bug you were seeing and how this fix corrects that? Thanks!

jpietrangelo:
Katie, I'm sorry, but I thought the comment made it clear. There was no "bug" on testbed.html. It's just that the HTML tag itself was missing. So I added it. As for the VirtualMachine.js file, I was trying to figure out why the Sample button in the testbed.com file does not produce the desired result when run (ie., it throws an error in the console, does not open a separate window, does not run any code, etc.) One of the things I figured out is that something is supposed to be removing \r and \n from the base64 code that's generated when reading the sample.py file. So I walked it back and thought that might be what the regex code in line #2 of VirtualMachine.js is supposed to do. But then I noticed that the search parameters were not escaped, which I believe they need to be. So I added the escape backslash to that line.

Sadly, I still haven't figured out why testbed.html does not work properly.

Hope that answers your question.

glasnt:
Hi @jpietrangelo. Thanks for that. 

The HTML tag change isn't required for HTML5, so while correct, it doesn't "fix" anything. 

The main issue with the bug you're seeing in the VirtualMachine.js is more pressing. Can you let me know what operating system, and browser you are using to get this error?

Ideally what I was looking for is: "I found this bug, I made this change, and this bug is now fixed". If this PR doesn't fix the bug you described, then it won't be merged.

jpietrangelo:
macOS Sierra (10.12.1)
Firefox
Safari

The error you get when pressing the "Run sample.py" button is:
InvalidCharacterError: String contains an invalid character

freakboy3742:
I'm going to close this - I haven't seen any other reports of this problem, including with people running macOS Sierra; I'm going to guess it was a transient problem with a particular revision. 

If you're still able to reproduce this problem, please let us know.

PR #391, opened 2016-12-04
Str.__mod__  

jstoebel:
fixes several tests for str.__mod__


jstoebel:
I've also begun writing additional tests for this. I just realized that I pushed up the beginnings of a these new tests. If you prefer, you can wait on merging until the tests are done as well. Or not, your call.

freakboy3742:
These look good; I'll wait until there's a bit more substance before merging.

jstoebel:
well it turns out the regex isn't as comprehensive as I once thought. I generated a combination of all of the possible specifiers (most of course are nonsense and will throw an exception but they need to be handled). [Most aren't caught by the regex](https://regex101.com/r/hysScU/1). The regex will need some tweaking to handle all of the corner cases.

jstoebel:
I am officially stumped on this. If anyone wants to offer suggestions I am all ears.

glasnt:
@jstoebel Do you have any updated code from your latest commit on this branch? At the moment all the tests just `pass`.

jstoebel:
not at the moment no. Would it be useful if I worked on just the tests and submitted those? I don't have a solution for the actual problem.

jstoebel:
ok I am going to attempt a different approach. Brython has abandoned the regex approach opting instead to loop over each character in the string to determine what action to take. I am going to attempt this approach as well.

jstoebel:
This has proved a tricky one, but I think I am close to having a working draft to share soon.

jstoebel:
FYI I am still working on polishing up the tests and cleaning things up a bit so they are more comprehensible but this is a rough draft. 

One of the biggest sticking points is [this](https://github.com/pybee/batavia/issues/399)

jstoebel:
FYI, I provided a way for tests to optionally skip certain parts of the transformation output. (see [issue 405](https://github.com/pybee/batavia/issues/405)). I blogged about this change [in a blog post](http://www.jstoebel.com/handling-test-output-in-batavia/)

tests/datatypes/test_str.py
```diff
@@ -195,6 +197,378 @@ def test_startswith_tuple(self):
...
+            and this is platform specific. currently, Batavia is not enforcing any
+            kind of upper bound.
+            """
+            values = [
+                32,
+                100,
+                -1,
+                '"s"',
+                '"spam"'
+            ]
```
freakboy3742: For internal purposes, it would be helpful to annotate why these particular cases are interesting (i.e, space, 'd', and three invalid values)

tests/datatypes/test_str.py
```diff
@@ -195,6 +197,378 @@ def test_startswith_tuple(self):
...
+            self.assertCodeExecution(tests, js_cleaner=js_cleaner, py_cleaner=py_cleaner)
+
+        @transforms(
+            js_bool=False,
+            decimal=False,
+            float_exp=False,
+            memory_ref=False
+        )
+        def test_precision(self, js_cleaner, py_cleaner):
+            percisions = ('.5', '.21')
```
freakboy3742: Spelling

jstoebel:
Thanks for all of this feedback. I'll get started on it!

jstoebel:
Thanks for this merge! There are still a handful of items requested by @freakboy3742 that I am still working on:

- [ ] Tests for `PYCleaner` and `JSCleaner`
- [] explicit attribute assignment in constructors for `JSCleaner` and `PYCleaner` (don't use `locals()`!)
- [] spelling fix (see last note from @freakboy3742 above)

I'll do a separate PR for these fixes.

And then of course I need to implement key word arguments...

awildtechno:
I'm having an issue with my floating points being truncated while testing the string formatter. Based on #405, it seems this issue should be fixed by your commit, but I don't see any examples of how to set up the test so that it's formatted correctly. Point me in the right direction?

martica:
You need to disable some part of the cleaners? I think you add @transforms() as a decorator on your test method. There are some examples of using @transforms to disable various cleaners in the tests.

PR #392, opened 2016-12-05
Implement Bool.__getitem__

rixx:


glasnt:
Hi @rixx! This PR needs to be rebased to include a fix to the CI configurations ✨

glasnt:
I've updated the CircleCI config on your branch.

rixx:
Woop woop, thanks, @glasnt! (First message of yours was at 1am local time, so I didn't see it until now). I'll rebase/cherry-pick, then we should be good.

PR #395, opened 2016-12-09
fixed byte_LOAD_ATTR

lucantrop:
Made it so `byte_LOAD_ATTR` properly handles `__getattribute__` and `__getattr__`.

### DIFF
--- a/batavia/VirtualMachine.js
+++ b/batavia/VirtualMachine.js
@@ -1327,28 +1327,13 @@ VirtualMachine.prototype.byte_COMPARE_OP = function(opnum) {
 }
 
 VirtualMachine.prototype.byte_LOAD_ATTR = function(attr) {
-    var PyObject = require('./core/types/Object')
     var obj = this.pop()
     var val
     if (obj.__getattribute__ === undefined) {
         // No __getattribute__(), so it's a native object.
         val = native.getattr(obj, attr)
     } else {
-        if (obj.__class__ !== undefined) {
-            var getattro = native.getattr_raw(obj.__class__.$pyclass.prototype,
-                '__getattribute__')
-
-            // if class of object has getattribute method,
-            // call that, otherwise, call
-            // object.__getattribute__
-            if (getattro !== undefined && getattro.__get__ !== undefined) {
-                val = getattro.__get__(obj).__call__(attr)
-            } else {
-                val = PyObject.__class__.__getattribute__(obj, attr)
-            }
-        } else {
-            val = obj.__getattribute__(attr)
-        }
+        val = native.getattr_py(obj, attr)
     }
 
     this.push(val)
--- a/batavia/VirtualMachine.js
+++ b/batavia/VirtualMachine.js
@@ -1115,7 +1115,9 @@ VirtualMachine.prototype.byte_LOAD_NAME = function(name) {
         val = frame.f_builtins[name]
         // Functions loaded from builtins need to be bound to this VM.
         if (val instanceof Function) {
+            var doc = val.__doc__
             val = val.bind(this)
+            val.__doc__ = doc
         }
     } else {
         throw new builtins.NameError.$pyclass("name '" + name + "' is not defined")
@@ -1161,7 +1163,9 @@ VirtualMachine.prototype.byte_LOAD_GLOBAL = function(name) {
         val = this.frame.f_builtins[name]
         // Functions loaded from builtins need to be bound to this VM.
         if (val instanceof Function) {
+            var doc = val.__doc__
             val = val.bind(this)
+            val.__doc__ = doc
         }
     } else {
         throw new builtins.NameError.$pyclass("name '" + name + "' is not defined")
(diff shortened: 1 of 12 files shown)


batavia/VirtualMachine.js
```diff
@@ -1300,15 +1300,26 @@ VirtualMachine.prototype.byte_COMPARE_OP = function(opnum) {
...
-            throw new builtins.AttributeError(
-                "'" + obj.__class__.__name__ + "' object has no attribute '" + attr + "'"
-            );
+            if (obj.__getattr__ === undefined) {
+                throw new builtins.AttributeError(
+                    "'" + obj.__class__.__name__ + "' object has no attribute '" + attr + "'"
+                );
+            } else if (obj.__getattr__ instanceof Function) {
+                val = obj.__getattr__(attr);
+            } else if (obj.__getattr__ instanceof types.Function) { // Shouldn't these already be methods?
```
freakboy3742: No - the Function is only bound to the instance as part of the __getattr__ process. The class has a single, common `types.Function` instance; each object instance gets a separate types.Method instance. In theory, this *could* be cached after first access, or at time of object construction; but Method is such a lightweight class it doesn't really matter.

tests/structures/test_class.py
```diff
@@ -46,6 +46,23 @@ def __str__(self):
             print('Done.')
             """, run_in_function=False)
 
+    # @unittest.expectedFailure
```
freakboy3742: This comment can be removed, too.

tests/structures/test_class.py
```diff
@@ -46,6 +46,23 @@ def __str__(self):
...
+                def __getattr__(self, attr):
+                    self.y = 42
+                    return y+1
+
+            obj = MyClass()
+            print(obj.x)
+            print(obj.fail)
+            print(obj.y)
+        """)
+
```
freakboy3742: This test will pass; but it's not a good test. If you actually run this, you'll get exceptions because the `y` on line 58 is an unknown variable. The error will be consistent between Python and Batavia, so the test passes; but it's not fully testing like you think it is.

It also doesn't test `__getattribute__`, which was the major contribution of this patch.

lucantrop:
Ah, of course, I should've checked in python before actually adding the tests :D

batavia/VirtualMachine.js
```diff
@@ -1300,15 +1300,26 @@ VirtualMachine.prototype.byte_COMPARE_OP = function(opnum) {
 VirtualMachine.prototype.byte_LOAD_ATTR = function(attr) {
     var obj = this.pop();
     var val;
-    if (obj.__getattr__ === undefined) {
+
+    if (obj.__getattribute__ === undefined) {
```
shaib: This appears to be wrong -- we should be checking for `__getattr__` and `__getattribute__` on the class, not on the instance:
```python
>>> class A:
...     pass
... 
>>> a = A()
>>> a.__getattr__ = lambda x:x
>>> 
>>> a.asdasdsad
  File "<stdin>", line 1, in <module>
AttributeError: 'A' object has no attribute 'asdasdsad'
>>> A.__getattr__ = lambda instance,x:x
>>> a.asdasdsad
'asdasdsad'
```

tests/structures/test_class.py
shaib: I think that, in view of the current implementation and that this is Javascript, you should also add tests in the spirit of my negative example above (setting `__getattr__` and `__getattribute__` on the instance to verify that it doesn't work).

lucantrop:
Correct me if I'm wrong, but to do anything _useful._ (including a decorator protocol) with user-defined `__getattribute__` and `__getattr__`, we'd need to add `__dict__` (and/or `object.__getattribute__` for avoiding recursion with `__getattribute__`), which probably means changing the `PyObject` type.
For now, I definitely need to fix what @shaib pointed out, and I'll try to make the tests more robust in some other way.

shaib:
Hi, @lucantrop, are you still working on this? mind a little collaboration?

lucantrop:
Yeah, sure, I was just going to ask for some help. In the testbed, the following code:
```
class A:
	pass

A.__getattribute__ = lambda instance, x: x

a = A()
print("class "+A.attr)
print("object "+a.attr)
```

prints:
```
class attr
  File "/tmp/tmp57ctpx0f", line 8, in <module>
AttributeError: 'A' object has no attribute 'attr'
```
which indicates that the object didn't inherit everything it should've from the class.
Also, 
```
class A:
    def foo(self):
        return 'bar'
a = A()
print(a.foo())
print(A.foo(a))
```
prints 
```
bar
  File "/tmp/tmp6tunw14q", line 6, in <module>
AttributeError: 'A' object has no attribute 'foo'
```
I suspect the problem lies somewhere in the `make_class` function in `VirtualMachine.js`, but, that's as far as I went into it.
On the other hand, I started implementing the `object` builtin, which is, as I said in a previous comment, needed in order to use `__getattribute__`. I'll push that part of the code when I get it working properly.

lucantrop:
It mostly works like cpython now, but there are a few fixes left to do.

lucantrop:
@freakboy3742 this is (finally :smile:) finished, so you can review it again.

PR #402, opened 2016-12-26
Add a conversion step for the README when publishing to npm.

freakboy3742:


glasnt:
pandoc needs to be installed on the machines before it can be used

something like this added to `circle.yml`

```yaml
dependencies:
  pre:
    - sudo apt-get update; sudo apt-get install pandoc 
```

Probably also with the `parallel` flag. See [doc](https://circleci.com/docs/installing-custom-software/)

mrPain89:
When I follow the getting started guide, `npm install` fails because pandoc is not installed. I could install it manually but I think it's still an issue because it's not explained that we use CircleCI to solve dependencies.
I have a haskell environment already set up and used `cabal install pandoc` to solve my dependency, however, would this be possible to use another specific program to convert the README.rst to README.md? I checked on pip and npm and there's not efficient tool to do this task.

freakboy3742:
It appears this is a problem caused by recent versions of npm changing the interpretation of "prepublish" - v3 didn't do a prepublish as part of an install; npm v4 does. The conversion to md is only required as part of uploading the package to NPM; it isn't required for general usage, so the fix here will be to correct the npm targets.

freakboy3742:
This should now be corrected - the npm configuration now uses the prebuild and prepublishOnly targets, instead of the prepublish target.

PR #407, opened 2017-01-05
Accept and use commandline arguments to locate Ouroboros

kennethlove:
The `--source` argument was added to `compile_stdlib.py` but wasn't yet used to locate Ouroboros' source files.

In addition to that, `npm install` wasn't able to accept or use any commandline arguments for the source location of Ouroboros, so even if you provided it to `python compile_stdlib.py`, it wouldn't be reflected when you used `npm install`.

This PR also contains a change in `package.json` turning `prepublish` into `prepare` as `prepublish` isn't run for local, dev-style installs as of `npm` 4ish. If skipping this is the desired procedure (seems to make tests impossible to run), I'll happily revert.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -10,7 +10,7 @@ dependencies:
   pre:
     - sudo apt-get install pandoc
   override:
-    - sudo npm install -g npm
+    - npm install -g npm
     - npm install
 
 test:
--- a/circle.yml
+++ b/circle.yml
@@ -10,6 +10,7 @@ dependencies:
   pre:
     - sudo apt-get install pandoc
   override:
+    - npm install npm
     - npm install
 
 test:
@@ -19,7 +20,6 @@ test:
     - sudo service mysql stop || true
     - sudo service postgresql stop || true
   override:
-    - npm install install
     - "pip install pytest":
         parallel: true
     - "pip install git+https://github.com/pybee/pytest-circleci":
--- a/circle.yml
+++ b/circle.yml
@@ -10,7 +10,7 @@ dependencies:
   pre:
     - sudo apt-get install pandoc
   override:
-    - npm install npm
+    - sudo npm install -g npm
     - npm install
 
 test:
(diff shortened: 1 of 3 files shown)


swenson:
It looks like tests are failing on CircleCI. I would suspect that we are using npm <= 3, since Node was last updated a few months ago in the CircleCI base image.

In general, I'm okay with moving onto npm 4 (yay future!), but this will require us to wait for CircleCI, or to manually update npm in our `circle.yml`. You can try to take a stab at updating `circle.yml` to update npm.

kennethlove:
So, from `README.md`:

> Install Node.js. You must have a recent version of Node; we do our testing using v6.9.1.

swenson:
Does that mean we should update the README as well to indicate that we need a newer version of npm than the default with `6.9.1`? I thought `6.9.1` was released before npm 4?

kennethlove:
I think I'm just confused on those two since Node and npm aren't the same thing. So, ignore that.

The CircleCI upgrade of `npm` seemed to work (no errors about the upgrade) but it didn't cause the tests to pass as it seems to be missing `./stdlib`. Not sure what to change there.

Unfortunate that you can't use modern Node/npm for local development...

swenson:
Hmm. I think the `npm` upgrade needs to be in the `dependencies` section, before `npm install`, to have the appropriate effect, right?

kennethlove:
AFAIK, it's fine where it is in the `test` section, but, I could be wrong on that. Happy to try it the other way.

swenson:
It still looks like the `stdlib` isn't getting built. Maybe try `sudo npm install -g npm` as the warning suggests.

freakboy3742:
I think the fix is actually slightly different, and a lot less invasive. We recently renamed the npm package, but I forgot to update the directory reference in `compile_stdlib.py`.

freakboy3742:
See #408 for a simpler approach (plus a couple of other cleanups)

freakboy3742:
Closing in favour of #408, which has now been merged.

PR #409, opened 2017-01-06
Tests for str.startswith

kennethlove:
Tests to cover basic functionality of `str.startswith`.

### DIFF
--- a/batavia/types/Str.js
+++ b/batavia/types/Str.js
@@ -623,7 +623,17 @@ Str.prototype.encode = function(encoding, errors) {
 }
 
 Str.prototype.startswith = function (str) {
-    return this.slice(0, str.length) === str;
+    var types = require('../types');
+
+    if (str !== None) {
+		if (types.isinstance(str, [types.Str])) {
+			return this.slice(0, str.length) === str;
+		} else {
+			throw new exceptions.TypeError(
+				"TypeError: startswith first arg must be str or a tuple of str, not " + type_name(str)
+			);
+		}
+	}
 }
 
 Str.prototype.endswith = function (str) {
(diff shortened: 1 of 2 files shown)


tests/datatypes/test_str.py
```diff
@@ -109,6 +109,42 @@ def test_slice(self):
...
+                print('BeeWare starts with BeeWare-shaves-yaks')
+            else:
+                print('BeeWare does not start with BeeWare-shaves-yaks')
+            # Empty string
+            if 'BeeWare'.startswith(''):
+                print('BeeWare starts with an empty string')
+            else:
+                print('BeeWare does not start with an empty string')
+            # Invalid type
+            if 'BeeWare'.startswith(5):
```
freakboy3742: ...and, this, boys and girls, is why we have CI. Of course - this one should raise an exception, not return False. So:

1) the patch needs to use a try-catch here to make sure the right exception is thrown; on throwing an exception, the test passes.

2) Either the test needs o be marked as an `@expectedFailure`, or the implementation needs a type check to throw the right exception.

kennethlove:
Appropriate try/except chunks added in.

This doesn't properly test the error messages yet. It also needs tests for multiple arguments, no arguments, and a tuple of possibilities. Not sure how to implement each of these.

batavia/types/Str.js
```diff
@@ -623,7 +623,17 @@ Str.prototype.encode = function(encoding, errors) {
 }
 
 Str.prototype.startswith = function (str) {
-    return this.slice(0, str.length) === str;
+    var types = require('../types');
+
+    if (str !== None) {
+		if (types.isinstance(str, [types.Str])) {
```
glasnt: Your tabs are running away with you here :)

glasnt:
> This doesn't properly test the error messages yet. It also needs tests for multiple arguments, no arguments, and a tuple of possibilities. Not sure how to implement each of these.

Could you stub these functions out with a TODO and an `@expectedFailure`? That will be helpful for the next person :)

PR #416, opened 2017-01-28
added pandoc pre-requisite

mrPain89:
Added a instruction to install pandoc pre-requisite. Maybe we should also write the command for debian-based systems installation of the package.


freakboy3742:
Thanks for the PR; but as noted elsewhere, we've fixed this by using different targets in the npm configuration.

PR #418, opened 2017-01-30
fix error output in noargs for chr

danisharsalan:
An error in chr.js was not being printed in the correct format.


freakboy3742:
Thanks for the patch - can you clarify what problem you were trying to fix here? Batavia has an extensive test suite to verify that the right error messages are being output, which has been picked up by CI. The fact that your patch doesn't include any test modifications is also a red flag. 

What test case were you trying to fix?

PR #419, opened 2017-02-01
Str.__mod__  

jstoebel:
this implements key word arguments for `str.__mod__`

I did this by creating a function called `Args` which abstracts away the fetching of arguments.

One place I wanted to call attention is the change I made to `KeyError`. KeyError returns something like `Key Error: '\\'spam\\''` when it should just return `KeyError: 'spam'`. I'm not quite sure the best way to address this.

### DIFF
--- a/batavia/types/StrUtils.js
+++ b/batavia/types/StrUtils.js
@@ -4,6 +4,8 @@ var BigNumber = require('bignumber.js').BigNumber
 
 function _substitute(format, args){
     var types = require('../types');
+
+    // INITIAL SETUP
     // does this conversion use key word or sequential args?
     var kwRe = /%\(/;
     var usesKwargs = (format.match(kwRe) !== null)
@@ -28,8 +30,9 @@ function _substitute(format, args){
         this.getArg = function(key){
             // get the next arg.
             // if using sequental args, arg is shifted and returned
-            // if using kwargs, arg is simply returned.
+            // if using kwargs, get value of key
             if (usesKwargs){
+                var result = this.collection.__getitem__(key);
                 return this.collection.__getitem__(key);
             } else {
                 return this.remainingArgs.shift();
@@ -689,6 +692,8 @@ function _substitute(format, args){
             if (err.message === 'illegal character') {
               var charAsHex = nextChar.charCodeAt(0).toString(16)
               throw new exceptions.ValueError(`unsupported format character '${nextChar}' (0x${charAsHex}) at index ${charIndex + index + 1}`)
+            } else if (err.name == "KeyError") {
+                throw new exceptions.KeyError(err.msg)
             } else {
               //its some other error
               throw err
--- a/batavia/types/StrUtils.js
+++ b/batavia/types/StrUtils.js
@@ -84,7 +84,6 @@ function _substitute(format, args){
             // if its '*' there needs to be at least 2 args left (one for the * and another for the conversion)
             // if its a conversion there needs to be atleast one left
 
-        // this.remainingArgs = args.slice();  // as args are needed, shift them from this array, then return what's left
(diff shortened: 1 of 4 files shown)


jstoebel:
The following test is failing and I'm not sure why. I don't _think_ it has anything to do with my changes (famous last words I know). Any insights?

```
FAIL: test_subclass (tests.structures.test_class.ClassTests)
  File "/Users/stoebelj/Dropbox/batavia/tests/structures/test_class.py", line 93, in test_subclass
    """)
  File "/Users/stoebelj/Dropbox/batavia/tests/utils.py", line 444, in assertCodeExecution
AssertionError: 'Myobject instance 37 42\n37\n74\n42\n1554\nDone.\n' != 'Myobject instance 37, 42\n37\n74\n42\n1554\nDone.\n'
- Myobject instance 37 42
+ Myobject instance 37, 42
?                     +
  37
  74
  42
  1554
  Done.
 : Global context

```

jstoebel:
Just checking in. Is there anything else this PR needs? I think its ready to go.

tests/datatypes/test_str.py
```diff
@@ -482,6 +500,90 @@ def test_bogus_specifier(self, js_cleaner, py_cleaner):
...
+            js_bool = False,
+            decimal = False,
+            float_exp = False,
+            memory_ref = False
+        )
+        def test_wildcard_with_kwargs(self, js_cleaner, py_cleaner):
+
+            tests = adjust("""
+                print(">>> '%*(spam)s' % {'spam': 'eggs'}")
+                print('%*s' % {'spam': 'eggs'})
```
freakboy3742: Not sure which of these you're intending to test here, but the test case doesn't match the debug line.

batavia/types/StrUtils.js
```diff
@@ -3,9 +3,53 @@ var type_name = require('../core').type_name;
...
+    function Args(args, usesKwargs) {
+        // represents either sequential or key word arguments.
+        // functions that need to grab arguments interface with this object
+        if (usesKwargs){
+            this.collection = args[0]
+        } else{
+            this.remainingArgs = args.slice(); // remaining args
+        }
+
+        this.getArg = function(key){
```
freakboy3742: Ideally this would be added to the prototype, rather than an attribute on the instance.

jstoebel:
Thanks I'll get started on these. You mentioned formatting issues. Could you provide more guidance on what these are? You don't need to go through and mark them all, but if there are any conventions I should follow I can check for them.

freakboy3742:
The best guidance I can give you on code formatting is to follow the general principles of spacing that is enforced by PEP8, but with 2 additions:

* if/for etc aren't functions - so `if (condition)`, not `if(condition)`
* K&R style braces (i.e., `if (condition) {`, with the closing brace at the same depth as the opening if) 

I want to set up some ESLint formatting rules so that these formatting checks can be applied automatically - but (1) I need to set that up, and (2) there's the task of fixing all the existing formatting problems in the code that I've ignored because it's too labor intensive to point them all out manually.

jstoebel:
Thanks, I'll work on that. Just out of curiosity I ran eslint under the airbnb rules and got a BUNCH of errors some of which seem to conflict with the Python way of doing things (2 space indents for example) Should I just ignore the current eslint.rc for now?

freakboy3742:
For the moment, yes - ignore the errors. At the moment, there's a combination of "their rules are dumb" (like 2 space indents) and actual code format problems. Hence the first problem being getting our code format style sorted.

jstoebel:
I think I have cleaned up all of the formatting problems. Apologies in advance if I missed any!

freakboy3742:
I am, however, going to find a way to make your editor autocorrect "percision"... :-)

PR #425, opened 2017-02-18
Nested collections bugs

OndrejVicar:
I found this during implementation of `__eq__` for  dict. #46 

```
CPython:
>>> [[1, 2], [3, 4]] == [[1, 2], [3, 4]]
>>> True

Batavia:
>>> [[1, 2], [3, 4]] == [[1, 2], [3, 4]]
>>> False
```

I also added nested collections to sample data to test other collections (looks OK).


OndrejVicar:
Those new collections reavealed few other bugs. I will fix them first.

jstoebel:
great catch!

OndrejVicar:
Ok I resolved bugs with nested sets (it also resolved more tests for `__ge__` and `__le__`  as side effect).
But I'm not sure what to do it with nested dict that i've added to sample data. Theres problem with assertCodeExecution when it deals with those dicts:
```
>>> f = dict
>>> x = ((1, 2), (3, 4), (5, 6))
>>> f(x)
- {5: 6, 1: 2, 3: 4}
+ {1: 2, 3: 4, 5: 6}
```
This should pass the test.
Should i comment out nested dict from sample data and create new issue, so this PR can be merged (tests will pass)?

jstoebel:
@OndrejVicar I'll let a core member verify this, but I think the typical practice here is to mark tests as `@unittest.expectedFailure` as a way to let others know that this test doesn't pass now but should be circled back to later.

OndrejVicar:
Yea, looks like good solution thanks :). But there is little problem. Since dict order is "random", there will be unexpected success (which is marked as failed)
For example this test:
```
tests.builtins.test_ascii.BuiltinAsciiFunctionTests.test_dict
```

freakboy3742:
Making tests with `@expectedFailure` is definitely the way to go if a test is valid, but not currently passing. 

However, if the failure is just due to dictionary order, we have a workaround in place: you can define substitutions. If you look at the tests for sets (tests/datatypes/test_set.py), you can see an example - you define a dictionary of "normalised" output - the keys being the "normal" form, with the values being a list of alternatives that are equivalent to the normal form.

OndrejVicar:
Thanks. Still few more problems. Working on them.

OndrejVicar:
I can't figure this out.
```
CPython:
>>>x = ((1, 2), (3, 4)) 
>>> min(x)
>>> (1, 2)

Batavia:
>>>x = ((1, 2), (3, 4)) 
>>> min(x)
>>> (3, 4)
```

jstoebel:
@OndrejVicar If I'm reading this right, it points to an error in the Batavia implementation that test cases have failed to catch. To venture a guess maybe its a problem with `Tuple.__lt__`?

glasnt:
Hi @OndrejVicar. This looks like it's passing CI now, but you're going to have to rebase based on the changes we made in #427. There is also now linting checks, so these may need to be cleaned up on your code. Let us know if you have any issues.

batavia/builtins/max.js
```diff
@@ -39,7 +39,7 @@ function max(args, kwargs) {
     try {
         while (true) {
             var next = iterobj.__next__()
-            if (next.__gt__(max)) {
+            if (next.__gt__(max).valueOf()) {
```
OndrejVicar: I dont know, why this caused problem **only** with nested tuples... But it helps.

batavia/builtins/sorted.js
```diff
@@ -59,12 +59,16 @@ function sorted(args, kwargs) {
...
-                return validatedInput['smaller']
+            // Order of conditions does matter here.
+            // Because if we get unorderable types, CPython gives always '<' in Exception:
+            // TypeError: unorderable types: str() < int()
+            if (a["key"].__lt__(b["key"])) {
+                return validatedInput["smaller"]
+            }
+            
+            if (a["key"].__gt__(b["key"])) {
+                return validatedInput["bigger"]
```
OndrejVicar: This was not caught earlier by tests only because __lt__ and __gt__ of Tuple was also broken (raising wrong exception message).

OndrejVicar:
@glasnt I think only CI problem now is resolved by #434. Otherwise, I think it's ready.

freakboy3742:
@OndrejVicar #434 has now been merged; can you sync the branch so we can see the full CI output?

OndrejVicar:
@freakboy3742 Done =)

PR #430, opened 2017-02-19
Implement __xor__ for tuples

treyhunner:
Very small step in the direction of #46


glasnt:
Every little bit helps, thanks!

PR #440, opened 2017-02-25
Add input validation so that test passes

thatandromeda:
`tests.datatypes.test_str.StrTests.test_startswith_multiple_args` no longer expected to fail. 🎉


thatandromeda:
This was originally written just to handle `test_startswith_multiple_args`, but it can and should also handle `test_startswith_no_args`, so now it does.

PR #443, opened 2017-02-25
Implemented the in operator, isupper() and islower() methods for strings

bararan:
What the title says: in operator as well as isupper() and islower() string methods have been added.


batavia/types/Str.js
```diff
@@ -198,7 +198,19 @@ Str.prototype.__ge__ = function(other) {
...
+    var types = require('../types')
+    if (types.isinstance(other, [
+        types.Bool, types.Int, types.Float,
+        types.List, types.Dict, types.Tuple,
+        types.Set, types.Bytearray, types.Bytes,
+        types.Type, types.Complex, types.NotImplementedType,
+        types.Range, types.Slice, types.FrozenSet
+
+    ])) {
+        throw new exceptions.TypeError.$pyclass('cannot search type ' + type_name(other) + ' in string')
```
wolfv: Isn't the correct message `TypeError: 'in <string>' requires string as left operand, not int`?

wolfv: Appropriately I think one could also just test for `types.isinstance(other, [types.String])`

freakboy3742: Yes - the error message is probably incorrect; the test proposed here follows an example from earlier in the suite for startswith. `in` should probably be added to the list of operators to be tested.

And yes, the condition could probably be simplified too. Fancy making a pull request? :-)

bararan: I'll make the changes -both to the method and the unit test.
Thanks for the comments @wolfv

bararan: It should all be fine now.

PR #446, opened 2017-02-25
Add ability to have bool add bool, Adding a utils function to return actual function

graph1994:
Basically called __add__ in __iadd__ to fix adding booleans together, added a utils.js file to make sure the actual method is not overridden when it errors

### DIFF
--- a/batavia/types/utils.js
+++ b/batavia/types/utils.js
@@ -1,9 +1,9 @@
 var utils = {}
 
-utils.inplace_call = function(f, operand_str, this, other) {
+utils.inplace_call = function(f, operand_str, this_obj, other) {
     // Call the method named "f" with argument "other"; if a type error is raised, throw a different type error
     try {
-        return this[f](other)
+        return this_obj[f](other)
     } catch (error) {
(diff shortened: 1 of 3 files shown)


batavia/types/utils.js
```diff
@@ -0,0 +1,19 @@
...
+var utils = {}
+
+utils.inplace_call = function(f, operand_str, this_obj, other) {
+    // Call the method named "f" with argument "other"; if a type error is raised, throw a different type error
+    try {
+        return this_obj[f](other)
+    } catch (error) {
+        if (error instanceof exceptions.TypeError.$pyclass) {
+            throw new exceptions.TypeError.$pyclass(
+                'unsupported operand type(s) for ' + operand_str + ": 'int' and '" + type_name(other) + "'")
```
wolfv: Shouldn't "'int'" be replaced with `type_name(this_obj)`? For the `Bool` case the resulting error message looks wrong to me.

graph1994: FIXED in #449

wolfv: awesome

PR #448, opened 2017-02-26
Now bool() properly handles __len__.

thatandromeda:
(this page of unkind statements about JavaScript intentionally left blank)

### DIFF
--- a/batavia/builtins/bool.js
+++ b/batavia/builtins/bool.js
@@ -1,4 +1,5 @@
 var exceptions = require('../core').exceptions
+var callables = require('../core').callables
 var types = require('../types')
 
 function bool(args, kwargs) {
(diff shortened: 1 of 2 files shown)


batavia/builtins/bool.js
```diff
@@ -17,7 +18,16 @@ function bool(args, kwargs) {
...
+        // args[0].__bool__, if it exists, is a batavia.types.Function.js,
+        // *not* a native Javascript function. Therefore we can't call it in
+        // the seemingly obvious way, with __bool__().
+        return callables.call_method(args[0], '__bool__', [])
+    // Python bool() checks for __bool__ and then, if __bool__ is not defined,
+    // for __len__. See https://docs.python.org/3.4/library/stdtypes.html#truth.
+    } else if (args[0].__len__) {
+        // valueOf() can return a string even when __len__ was set to an
+        // integer (as in builtins.test_bool.BoolTests.test_len_only).
+        return !!parseInt(callables.call_method(args[0], '__len__', []).valueOf())
```
thatandromeda: I thought about that and decided I was unsure. Mostly because I realized I don't actually trust people to have implemented `__len__` in such a way that it returns something integer-like, so if someone's written a __len__ that returns, say, "seventeen" or ⛔️ , I'd rather have something that I'm pretty sure will fail dramatically.

I didn't have the time at the sprint to think through all the angles on that question, so I'm open to other arguments :)

thatandromeda: Wait! The *correct* answer is that I should write some Python classes such that `__len__` returns "seventeen" or whatever, see what errrors they throw, write tests checking for that error, and make them pass. Coolio. I'll do that later today.

wolfv: btw. my answer was wrong

```
In [6]: class A:
    def __len__(self):
        return "12"
   ...:     

In [7]: a = A()

In [8]: bool(a)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-7af9bc0b6a67> in <module>()
----> 1 bool(a)

TypeError: 'str' object cannot be interpreted as an integer
```

PR #449, opened 2017-02-26
Clean up old __call_type_error_str__ in INT, Fix utils.js error

graph1994:
### CHANGES:
* Int.js was using __call_type_error_str__  which could be replaced by the new utils method.
* Fixed inplace_call method to return the type of this in error. 
* ~~Added lint to scripts of package.json so that `npm run lint` can be used~~
* Fixed unexpected success test in test_bool.py


glasnt:
Hi! There are some unexpected successes that need to be removed from `test_add_frozenset` because they now work. 

I'd also prefer if you kept this change to just the INT changes, and put the package.json change in a separate PR, please.

graph1994:
Okay, removed the package.json change, and hopefully fixed all of the unexpected successes caused by adding the __iadd__ for bool earlier :(

glasnt:
Thanks @graph1994! If you throw the package.json changes into a new PR and tag me, that'd be awesome

PR #451, opened 2017-02-27
Added lint command to package.json so now npm run lint can be used

graph1994:
CHANGE:
* Added lint script to package.json so now `npm run lint` can be used

@glasnt


glasnt:
Thanks!

PR #454, opened 2017-03-04
Implemented lstrip, rstrip and strip methods for string

bararan:
lstrip, rstrip and strip string methods added.

### DIFF
--- a/batavia/types/Str.js
+++ b/batavia/types/Str.js
@@ -550,7 +550,7 @@ Str.prototype.rstrip = function() {
             throw new exceptions.TypeError.$pyclass('rstrip arg must be None or str')
         }
         var result = this.valueOf()
-        while (charsToTrim.indexOf(result[result.length-1]) > -1) {
+        while (charsToTrim.indexOf(result[result.length - 1]) > -1) {
             result = result.slice(0, -1)
         }
         return result
@@ -573,7 +573,7 @@ Str.prototype.strip = function() {
         while (charsToTrim.indexOf(result[0]) > -1) {
             result = result.slice(1)
         }
-        while (charsToTrim.indexOf(result[result.length-1]) > -1) {
+        while (charsToTrim.indexOf(result[result.length - 1]) > -1) {
             result = result.slice(0, -1)
         }
         return result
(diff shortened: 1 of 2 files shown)


bararan:
Just noticed a bug in the unit tests (an extra indent that prevented some tests from running). Working on fixing that now.

batavia/types/Str.js
```diff
@@ -519,6 +519,69 @@ Str.prototype.encode = function(encoding, errors) {
...
+
+    if (arguments.length === 0) {
+        return this.valueOf().trimLeft()
+    } else if (arguments.length === 1) {
+        var charsToTrim = arguments[0]
+        if (!types.isinstance(charsToTrim, [types.Str])) {
+            throw new exceptions.TypeError.$pyclass('lstrip arg must be None or str')
+        }
+        var result = this.valueOf()
+        while (charsToTrim.indexOf(result[0]) > -1) {
```
wolfv: maybe it would be better to count up, and then slice? could be a little more performant.
ie.:

```js
var i = 0;
while (charsToTrim.indexOf(result[i]) > -1) {
    i++;
}
result = result.slice(i);
```

bararan: Done. Thanks for the review!

wolfv: great work! :tada:

tests/datatypes/test_str.py
```diff
@@ -301,57 +301,288 @@ def test_isupper(self):
...
+                print('12!3_4 is not lowercase')
+
+            print('done.')
+        """)
+
+    def test_lstrip(self):
+        # No argument passed, strip whitespace only
+        self.assertCodeExecution("""
+            x = ' Bee'
+            print(s.lstrip())
```
freakboy3742: I think you'll find this is a false positive test pass. The string is `x`, the test is on `s`. Both Batavia and Python will have the same result - but it will be a `NameError`, not the result of `s.lstrip()`.

On a related note - Batavia tests have a bit of an overhead associated with starting an `assertCodeExecution`, so it's advantageous to run a bunch of tests in one code block if possible (like `test_islower` does), rather than having a lot of independent test executions.

bararan: Oops...That was rather dumb, sorry. 

Now fixed the false positives & put the unit tests in one code block (and it did get remarkably faster!)

Thanks for the comment!

PR #458, opened 2017-03-09
Implemented addition and subtraction for booleans with complex numbers

kunparekh18:


kunparekh18:
Did not remove ```test_add_complex``` from the ```not_implemented``` tests in ```test_bool.py``` , causing an Unexpected Success on ```tests.datatypes.test_bool.InplaceBoolOperationTests.test_add_complex``` which leads to CircleCI tests failing. Should be fixed by removing said line.

kunparekh18:
That is because the ```__isub__()``` function for Booleans throws a NotImplementedError exception. Replacing the line with ```return utils.inplace_call('__sub__', '-=', this, other)``` solves the problem and passes all ```tests.datatypes.test_bool.InplaceBoolOperationTests.test_subtract_*``` tests. I'll send in a PR with this change once I finish implementing a few more functions.

PR #459, opened 2017-03-09
Add tuple multiplication

wolfv:


batavia/types/Tuple.js
```diff
@@ -256,7 +256,9 @@ Tuple.prototype.__mul__ = function(other) {
     if (types.isinstance(other, types.Int)) {
         var result = new Tuple()
         for (var i = 0; i < other.valueOf(); i++) {
-            result.extend(this)
+            for (var j = 0; j < this.length; j++) {
+                result.push(this[j])
+            }
```
freakboy3742: This looks great, and it clearly works - but I think this is a case where the problem should be fixed a little further down stream. Calling the Python `extend()` on a tuple *should* be the right answer here; but it appears extend hasn't been implemented. Rather than duplicating what will ultimately be the logic for extend, we should be adding an extend method to Tuple.

wolfv: but there is no extend method on a tuple in Python :(

wolfv:
```
Python 3.5.2 (default, Sep 14 2016, 11:28:32) 
[GCC 6.2.1 20160901 (Red Hat 6.2.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = (1,2,3,4)
>>> a.extend
  File "<stdin>", line 1, in <module>
AttributeError: 'tuple' object has no attribute 'extend'
>>> a.extend((1,2,3,4))
  File "<stdin>", line 1, in <module>
AttributeError: 'tuple' object has no attribute 'extend'
```

freakboy3742: /me facepalms.

Of course. Ignore me :-)

PR #463, opened 2017-03-14
Add str.capitalize() and test

ioo11:


tests/datatypes/test_str.py
```diff
@@ -609,6 +609,22 @@ def test_strip(self):
 
             print('done.')
             """)
+        
+    def test_capitalize_no_args(self):
+        self.assertCodeExecution("""
+        x = 'abcde'
+        print(x.capicapitalize())
```
wolfv: intentional `capicapitalize`? :)

batavia/types/Str.js
```diff
@@ -654,6 +654,14 @@ Str.prototype.__hash__ = function() {
     return new types.Int(h)
 }
 
+Str.prototype.capitalize = function() {
+    if(arguments.length === 0){
```
wolfv: I think for code style there are some spaces required around the brackets, ie. `if (...) {`

tests/datatypes/test_str.py
ioo11: bug) thx

batavia/types/Str.js
ioo11: recently i was writing mostly on c#, so i have used the same style here. Thank you!

wolfv: great stuff!

tests/datatypes/test_str.py
```diff
@@ -609,6 +609,22 @@ def test_strip(self):
 
             print('done.')
             """)
+        
+    def test_capitalize_no_args(self):
+        self.assertCodeExecution("""
+        x = 'abcde'
```
freakboy3742: This is a great start, but it would be good to do a little more robust testing. At the very least, checking capital letters and whitespace are unchanged; but it would also be good to check ascii < 32 control characters, unicode characters, characters with accents, and so on. 

You don't necessarily have to have an implementation that *works* for all of those cases - but if you are adding the capitalise method, we'd like to know what cases remain to be solved. You can mark those cases with `@expectedFailure` in your test suite.

wolfv: On a similar note, I think this case is not covered:

```
>>> "ASDV'".capitalize()
"Asdv'"
>>> "ASsd aSda Asd".capitalize()
'Assd asda asd'
```

ioo11: Yes, you were right, capitalize() didn't work with CamelCase. Fixed

ioo11: @freakboy3742 Thank you, I have a little experience in tdd. Now I tried to make test more complete.

PR #464, opened 2017-03-14
Start to build support for complex number exponentiation

thatandromeda:
This takes care of a lot of the test cases up front, because the right
answer for most types of exponent is "throw an error". Supporting
boolean exponents is easy because under the hood they're 1 or 0.
Supporting other types may want an additional JS library rather than
reimplementing the wheel, so let's defer that for a later PR.


thatandromeda:
OK, those last two commits should take care of the issues that CI had with the first commit.

freakboy3742:
Looks like a bunch of unexpected successes in `BuiltinTwoargPowFunctionTests`.

PR #465, opened 2017-03-18
Added swapcase() for String (Refs #46)

bitdumper1:


tests/datatypes/test_str.py
```diff
@@ -164,6 +164,14 @@ def test_startswith(self):
             print('done.')
         """)
 
+    def test_swapcase(self):
+        self.assertCodeExecution("""
+            test = ["test", "Test", "This Is a Test", ""]
```
freakboy3742: It would be good to check a few non-ascii characters (punctuation, accented characters, and so on) - even if the answer is "they don't work", it clearly indicates what improvements are required, captured as a test.

PR #471, opened 2017-03-23
implemented Tuple.__getitem__ with stepped slices

oopsno:
This workaround uses exactly the same code from `List.__getitem__` for consistency.

A full-featured slice seems to need some further work on the VM or our object model to add magic method support into Batavia, e.g. `object.__index__`.


shaib:
This appears to have already been implemented.

Issue #452, opened 2017-02-28
Problem while running sample.py in Batavia testbed


Sanjeev-S:
**How to replicate the error**

Pull the latest branch and setup the project following the basic tutorial.
https://batavia.readthedocs.io/en/latest/intro/tutorial-0.html
https://batavia.readthedocs.io/en/latest/intro/tutorial-1.html

_Now launch localhost and run sample.py._ The console output ends with:

```
Traceback (most recent call last):
  File "sample.py", line 85, in <module>
  File "sample.py", line 79, in main
  File "sample.py", line 39, in try_builtins
TypeError: bad operand type for abs(): 'NoneType'
```

For reference:
https://drive.google.com/file/d/0B9j9MJrQhCt2STJYOVJlVHIzRmc/view?usp=sharing

freakboy3742:
This is expected behavior; the error is thrown to check that error handling works. If you look at the code, it's the result of invoking `abs(None)`. However, I agree that it is surprising; the example should be modified so it either doesn't throw an error, or the error is caught.

morrme:
so is the fix for the issue now just a change in the sample file?

freakboy3742:
@morrme Correct - the sample.py file should be modified so that it catches the error and prints an appropriate response.

morrme:
So you want to keep the abs(None) there but modify try_builtins() to catch the TypeError exception?

freakboy3742:
Correct.

morrme:
I looked at other code in the project to come up with a consistent error message, but it's up to you: 

```
 
def try_builtins():
    try:
        print('sum(0,1,2,3,4)', sum([0, 1, 2, 3, 4]))
        print('abs(-1)', abs(-1))
        print(bool.__doc__)
        print('min(1,2,3,4)', min([1, 2, 3, 4]))
        print('max(1,2,3,4)', max([1, 2, 3, 4]))
        print('all([True, True, False])', all([True, True, False]))
        print('any([True, True, False])', any([True, True, False]))
        print('bool("String")', bool("String"))
        print('bool(false)', bool(False))
        print('hex(14)', hex(14))
        print('oct(14)', oct(14))
        print('bin(14)', bin(14))
        print('divmod(5,2)', divmod(5, 2))
        print('pow(2, 3)', pow(2, 3))
        print('pow(2, 3, 3)', pow(2, 3, 3))

        print('abs(None)', abs(None))  # known failure (we can also edit this comment)
    
    except: Type Error:
[...]

PR #473, opened 2017-03-24
sample.py updated with exception handling

morrme:
Fixes issue #452 .

### DIFF
--- a/testserver/sample.py
+++ b/testserver/sample.py
@@ -19,27 +19,26 @@ def do_stuff(count, size=3):
         other.wiggle(i)
 
 def try_builtins():
-    try:
-        print('sum(0,1,2,3,4)', sum([0, 1, 2, 3, 4]))
-        print('abs(-1)', abs(-1))
-        print(bool.__doc__)
-        print('min(1,2,3,4)', min([1, 2, 3, 4]))
-        print('max(1,2,3,4)', max([1, 2, 3, 4]))
-        print('all([True, True, False])', all([True, True, False]))
-        print('any([True, True, False])', any([True, True, False]))
-        print('bool("String")', bool("String"))
-        print('bool(false)', bool(False))
-        print('hex(14)', hex(14))
-        print('oct(14)', oct(14))
-        print('bin(14)', bin(14))
-        print('divmod(5,2)', divmod(5, 2))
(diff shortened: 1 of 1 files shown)


testserver/sample.py
```diff
@@ -18,26 +18,28 @@ def do_stuff(count, size=3):
...
-    print('hex(14)', hex(14))
-    print('oct(14)', oct(14))
-    print('bin(14)', bin(14))
-    print('divmod(5,2)', divmod(5, 2))
-    print('pow(2, 3)', pow(2, 3))
-    print('pow(2, 3, 3)', pow(2, 3, 3))
-
-    print('abs(None)', abs(None))  # known failure
-
+    try:
```
freakboy3742: This `try` clause only needs to cover the statement that we *know* will raise an error - otherwise, it could mask other problems.

testserver/sample.py
```diff
@@ -18,26 +18,28 @@ def do_stuff(count, size=3):
...
+        print('hex(14)', hex(14))
+        print('oct(14)', oct(14))
+        print('bin(14)', bin(14))
+        print('divmod(5,2)', divmod(5, 2))
+        print('pow(2, 3)', pow(2, 3))
+        print('pow(2, 3, 3)', pow(2, 3, 3))
+
+        print('abs(None)', abs(None))  # known failure 
+    
+    except: TypeError:
```
freakboy3742: Um... Are you sure you tested this? This shouldn't be legal syntax.

testserver/sample.py
```diff
@@ -18,26 +18,28 @@ def do_stuff(count, size=3):
...
+        print "Type Error: Invalid Argument"
```
freakboy3742: How do you know that this error is an invalid argument? And again - did you test this? Because print is a function, not a statement in Python 3.

morrme: The error is: 

> TypeError: bad operand type for abs(): 'NoneType'

In the issue, I explained the choice of the term "Invalid Argument": I was not sure if you had a convention so I referenced other code in the repo. That was the reason for me asking if you had a specific error message you wanted to use.

morrme:
@freakboy3742 thank *you*!

PR #485, opened 2017-04-02
Added the Powers of Complex Numbers

EGJ1996:
It supports a complex number to the power of: complex number, int, float, bool, class, bytes, range, str, set, frozenset, dict.

### DIFF
--- a/batavia/types/Complex.js
+++ b/batavia/types/Complex.js
@@ -284,11 +284,16 @@ function powi(x, y) {
     if (x.real === 0 && x.imag === 0) {
         return new Complex(0, 0)
     }
-    if (x.real === 0 && (x.imag === 1 || x.imag === -1)) {
-        return powu(x, y)
-    }
     if (Number(y) >= MAX_INT) {
+        if (x.real === 0 && (x.imag === 1 || x.imag === -1)) {
+            return powu(x,y)
+        }
         if (Number(y) <= MAX_FLOAT) {
+            if (x.real === 0 && (x.imag === 1 || x.imag === -1)) {
+                throw new exceptions.ZeroDivisionError.$pyclass(
+                    '0.0 to a negative or complex power'
+                )
+            }
             throw new exceptions.OverflowError.$pyclass(
                 'complex exponentiation'
             )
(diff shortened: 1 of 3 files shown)


batavia/types/Complex.js
```diff
@@ -197,28 +198,151 @@ Complex.prototype.__abs__ = function() {
...
+}
+function powi(x, y) {
+    if (x.real === 0 && x.imag === 0) {
+        return new Complex(0, 0)
+    }
+    if (x.real === 0 && (x.imag === 1 || x.imag === -1)) {
+        return powu(x, y)
+    }
+    if (Number(y) >= MAX_INT) {
+        if (Number(y) <= MAX_FLOAT) {
```
swenson: This is always true except for `Infinity` (ditto below for `MIN_FLOAT` and `-Infinity`). So, it may not be necessary to have the `MIN_FLOAT` and `MAX_FLOAT` constants, and just check for infinities directly.

batavia/types/Complex.js
```diff
@@ -197,28 +198,151 @@ Complex.prototype.__abs__ = function() {
...
+    } else if (abs_bimag >= abs_breal) {
+        const ratio = b.real / b.imag
+        const denom = b.real * ratio + b.imag
+        r = new Complex((a.real * ratio + a.imag) / denom, (a.imag * ratio - a.real) / denom)
+    } else {
+        r = new Complex(NaN, NaN)
+    }
+    return r
+}
+function powu(x, y) {
```
swenson: This looks like it only works for positive integer `y`, but below it could be called with a positive real number?

batavia/types/Complex.js
EGJ1996: But I have to consider 2 cases: 1. the number is greater than MAX_INT and smaller than MAX_FLOAT. If this is the case display the message exponentiation error. Otherwise, display the error message: number to big to be converted to float.

swenson: Right, but `Number(y)` will just return `Infinity` if it is greater than`MAX_FLOAT`, so the comparison is not needed -- you can just check if it is infinity.

EGJ1996:
@swenson  I cannot understand why in the last test there are some test cases with two values for the base.

swenson:
The built-in `pow` function accepts three arguments: it computes `pow(a, b, m) = (a**b)%m` efficiently. This form only supports integers though.

martica:
Thanks for all your work on this @EGJ1996. I picked up @shaib's version of this and got it up to date and merged.

PR #488, opened 2017-04-06
Raise a TypeError when a non-iterable is passed to filter.

judy2k:


judy2k:
I need to sign my commits!

judy2k:
Nope, apparently I don't

PR #490, opened 2017-04-06
Raise a TypeError when a non-iterable is passed to filter.

judy2k:
When a non-iterable is passed to `filter`, it should raise a TypeError. Now it does.

Signed-off-by: Mark Smith <mark.smith@practicalpoetry.co.uk>


judy2k:
OMG this is the most exciting thing that has happened to me! 😸

PR #497, opened 2017-04-06
Enabling test_setattr in test_list.py

martharotter:
Fixes test test_set_attr by implementing a check type on native.js for setattr to complain if the type does not have that attribute
Removes expectedFailure decorator from this test


shaib:
LGTM

martica:
The allows the test in test_list.py to pass, but causes some failures elsewhere now. The requirements for this are slightly more complex it seems. I'm going to close out this change, but feel free to reopen it and see if it can be updated if you are interested.

PR #499, opened 2017-04-07
Starting to work on the compiler/parser

swenson:
Building on the work of many months ago, porting over the CPython parser to JavaScript.

It's super buggy now, but at least it doesn't crash with some trivial input.

I'm finding it hard to make progress without emulating the entire CPython API, so I'm debating how to proceed. It's a little hard to break into pieces.


phildini:
Hi @swenson ! It looks like this PR might be dead, so we're closing it for now. Feel free to re-open it if you'd like to continue. 😄

PR #504, opened 2017-04-08
Update implementation of Slice

rebkwok:
Add TypeErrors for incorrect number of parameters
Assign None instead of null if no parameter passed
Fix Slice __str__ output
Throw TypeErrors for invalid slice indicies in Str, List, Range and Tuple
Fix Tuple slice implementation to properly deal with step of 0
Add to datatype slice tests to test invalid values for step
Add to datatype slice tests to test reverse step 
Add to datatype tuple tests to test slice

Signed-off-by: Becky Smith <rebkwok@gmail.com>

### DIFF
--- a/batavia/builtins/slice.js
+++ b/batavia/builtins/slice.js
@@ -1,17 +1,30 @@
+var exceptions = require('../core').exceptions
 var types = require('../types')
+var None = require('../core').None
 
 function slice(args, kwargs) {
+    if (!args || args.length === 0 || args.length > 3) {
+        throw new exceptions.TypeError.$pyclass('slice expected at least 1 arguments, got 0')
+    }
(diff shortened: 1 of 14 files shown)


batavia/builtins/slice.js
```diff
@@ -1,17 +1,30 @@
+var exceptions = require('../core').exceptions
 var types = require('../types')
+var None = require('../core').None
 
 function slice(args, kwargs) {
+    if (!args || args.length === 0 || args.length > 3) {
+        throw new exceptions.TypeError.$pyclass('slice expected at least 1 arguments, got 0')
```
freakboy3742: This error will be wrong for `slice(1,2,3,4)`

PR #516, opened 2017-05-13
Implemented len() for int and float

krtk6160:


krtk6160:
@freakboy3742 Please take a look. :)

krtk6160:
Thanks so much! That was my first pull request ever. :)

hdz666:
Keep it up lad 👍

PR #517, opened 2017-05-16
Byte add binary operator

kartikadur:
Fixed the `__str__` function in Bytes.js from comparing str with ascii (int val) to comparing two ascii/unicode values. Added code to check addition for Bytes with Bool, Bytes, Dict, Float, Int, List, Set, Str, Tuple.


tests/datatypes/test_bytes.py
```diff
@@ -105,23 +105,23 @@ class BinaryBytesOperationTests(BinaryOperationTestCase, TranspileTestCase):
     data_type = 'bytes'
 
     not_implemented = [
-        'test_add_bool',
+        # 'test_add_bool',
```
swenson: I would just delete these lines rather than comment them out. We won't need to add them back in again hopefully :)

kartikadur: I making the changes now, so I had a question: is there a way to resubmit a pull request after commiting changes or do I need to create a new pull request everytime?

kartikadur:
changes made and resubmitted. let me know if there are any other changes required.

kartikadur:
awesome! thank you

PR #519, opened 2017-05-22
Allow indexing tuples with Booleans.

khorn:
Fixes a problem where using a boolean as a tuple index in python code would throw an exception.

### DIFF
--- a/batavia/types/Tuple.js
+++ b/batavia/types/Tuple.js
@@ -325,7 +325,11 @@ Tuple.prototype.__getitem__ = function(index) {
             }
         }
     } else if (types.isinstance(index, types.Bool)) {
-        return this[index ? 1 : 0];
+        if (index) {
+            return this[1]
+        } else {
+            return this[0]
+        }
     } else if (types.isinstance(index, types.Slice)) {
         var start, stop, step
         if (index.start === None) {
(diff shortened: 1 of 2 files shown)


batavia/types/Tuple.js
```diff
@@ -324,6 +324,8 @@ Tuple.prototype.__getitem__ = function(index) {
                 return this[idx]
             }
         }
+    } else if (types.isinstance(index, types.Bool)) {
+        return this[index ? 1 : 0];
```
swenson: Would it be possible to add a test that fails now, but that this code fixes? This way we don't accidentally break it again. :)

Something as simple as adding a test that checks `[1, 2][True] == 2` would be great.

khorn: Activated the relevant test.

swenson: Awesome!

batavia/types/Tuple.js
```diff
@@ -324,6 +324,12 @@ Tuple.prototype.__getitem__ = function(index) {
                 return this[idx]
             }
         }
+    } else if (types.isinstance(index, types.Bool)) {
+        if (index) {
+            return this[1]
```
swenson: Ah, so, it looks like this is failing if the list is too short to support these indices. We'll have to throw IndexError if the list is too short or empty.

PR #523, opened 2017-05-22
Implement bytearray on bool type

ilankham:
Implementation of the following behavior:

- bytearray(True) -> bytearray(b'\x00')

- bytearray(False) -> bytearray(b'')

with corresponding tests enabled and passed.


tests/builtins/test_bytearray.py
```diff
@@ -10,7 +10,7 @@ class BuiltinBytearrayFunctionTests(BuiltinFunctionTestCase, TranspileTestCase):
 
     not_implemented = [
         'test_noargs',
-        'test_bool',
+        # 'test_bool',
```
phildini: Would you mind just removing this line instead of leaving it commented out?

PR #526, opened 2017-05-22
fix: Added fix for lshift and rshift

shortdiv:
This removes an expected failure and makes rshift and lshift work with Booleans.

cc @toastdriven

### DIFF
--- a/batavia/types/Bool.js
+++ b/batavia/types/Bool.js
@@ -520,10 +520,10 @@ Bool.prototype.__lshift__ = function(other) {
         }
     } else if (types.isinstance(other, types.Int)) {
         if (other.valueOf() < 0) {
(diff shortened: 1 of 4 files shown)


batavia/types/Bool.js
```diff
@@ -515,6 +515,12 @@ Bool.prototype.__lshift__ = function(other) {
             return new types.Int(0)
         }
     } else if (types.isinstance(other, types.Int)) {
+        if(other.valueOf() < 0) {
```
swenson: The linter is *probably* going to yell at you about having spaces after `if` (here and below).

shortdiv: Thanks for catching that! Updating code...

glasnt:
I've just pushed a fix for the linting issues. Will merge when CI goes green

Issue #520, opened 2017-05-22
Batavia testserver appears to be broken in Safari 9.1.3


ilankham:
Here's the output I saw in the Console:

```
[Error] TypeError: data.slice is not a function. (In 'data.slice(0, 4)', 'data.slice' is undefined)
	run (batavia.js:25114)
	bootstrap (127.0.0.1:474)
	onload (127.0.0.1:495)
```

lschumm:
@ilankham Please provide more information on your issue. I've run it in Safari, and have not had a similar issue. Are you sure you've followed all the instructions verbatim?

ilankham:
After following http://batavia.readthedocs.io/en/latest/intro/tutorial-0.html to download and build Batavia, I was able to run the testserver using Django version 1.10 and execute Python code in Chrome 58.0.3029.110.

However, the above error occurred when accessing the testserver in Safari 9.1.3 and attempting to execute the Python code `print("Hello, World")`. In addition, no console output was generated.

lschumm:
Ah, I didn't read your Safari version. Safari 9 doesn't support the `.slice()` method; code should be changed to implement a workaround. Similar issue: https://github.com/crocodilertc/crocodile-rtc/issues/15.

ilankham:
This issue has been addressed in pull request #527

ilankham:
The testserver now appears to work as expected in Safari 9, so this issue is being closed.

PR #527, opened 2017-05-22
Add workaround for Safari 9 Support

ilankham:
This pull request is intended to resolve issue #520 by adding a workaround for Safari 9's lack of implementation for TypedArray.prototype.slice(), per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/slice

### DIFF
--- a/batavia/core/types/PYCFile.js
+++ b/batavia/core/types/PYCFile.js
@@ -5,6 +5,12 @@ var constants = require('../constants')
  *************************************************************************/
 
 var PYCFile = function(data) {
+
+    if (!Uint8Array.prototype.slice) {
+        Object.defineProperty(Uint8Array.prototype, 'slice', {
+            value: Array.prototype.slice
+        });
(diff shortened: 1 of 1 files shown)


batavia/core/types/PYCFile.js
```diff
@@ -5,6 +5,12 @@ var constants = require('../constants')
  *************************************************************************/
 
 var PYCFile = function(data) {
+
+    if (!Uint8Array.prototype.slice) {
+        Object.defineProperty(Uint8Array.prototype, 'slice', {
+            value: Array.prototype.slice
+        });
```
swenson: The linter might complain about this semicolon (I don't think we have them elsewhere.

glasnt:
@ilankham I've committed a linting fix to this PR, including an ignore for the native extension (that is the point of this PR)
I'll merge once it runs green

PR #531, opened 2017-05-23
Update mul, truediv, div, sub, and add functionality

kartikadur:
Updated Bytes.js to include multiply, divide, add, and substract operations between Bytes Object type and the following Object types: Bool, Bytes, Dict, Float, Int, List, None, Set, Str, Tuple.


kartikadur:
thank you!

PR #533, opened 2017-05-23
Update complex exponentiation

lschumm:
Complex.__pow__ now supports:
- Complex ** Bool (did before)
- Complex ** Int
- Complex ** Float
- Complex ** Complex

### DIFF
--- a/batavia/builtins/open.js
+++ b/batavia/builtins/open.js
@@ -1,8 +1,10 @@
 var exceptions = require('../core').exceptions
+var types = require('../types')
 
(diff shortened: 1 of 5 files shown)


shaib:
This has been implemented in a way which mimicks CPython much more closely in #753. Seeing the lack of reponse from original author, I think this can safely be closed.

martica:
We've got the complex exponentiation work from #485 to #753 to #794 in now so this PR is obsolete. Thanks for work on this, @lschumm.

PR #534, opened 2017-05-23
Lots of word-wrapping, and a more accurate sandbox tutorial

kjcole:
I believe markdown and reStructuredText are both designed to be readable in a 80-character width console window. So, I wrapped wherever I could.

Also, the tutorial for the sandbox described a different screen layout than what I was seeing. So, I changed the text to agree with the location of the sections as they currently appear on the screen.

The community "getting help" needed updating as well.

### DIFF
--- a/docs/community/community.rst
+++ b/docs/community/community.rst
@@ -5,9 +5,17 @@ Batavia is part of the `BeeWare suite`_. You can talk to the community through:
 
 * `@pybeeware on Twitter`_
 
-* The `BeeWare Users Mailing list`_, for questions about how to use the BeeWare suite.
+* See the `Getting Help`_ page for information on where to ask
+  questions regarding the use of the BeeWare suite, as well as links
+  to discussions the development of new features in the BeeWare suite,
+  and ideas for new tools for the suite.
 
-* The `BeeWare Developers Mailing list`_, for discussing the development of new features in the BeeWare suite, and ideas for new tools for the suite.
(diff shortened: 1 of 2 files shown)


phildini:
Hey there! Sorry it's taken so long to reply on this, and thank you so much for your submission.

I know this might seem arbitrary, but we'd like to see you change this PR to only include the text changes and additions you're contributing. To be more specific, the line length fixes are not something we're super interested in right now, and make this PR hard to review for the content. Another reason we're less excited about the line length fixes is that it makes the git history muddier, and clear git history becomes crucial on a project of this size.

We definitely bear responsibility for not making our styleguide clearer, and we're excited for your content changes once the line length fixes are dropped.

docs/community/community.rst
```diff
@@ -5,9 +5,17 @@ Batavia is part of the `BeeWare suite`_. You can talk to the community through:
...
-* The `BeeWare Users Mailing list`_, for questions about how to use the BeeWare suite.
+* See the `Getting Help`_ page for information on where to ask
+  questions regarding the use of the BeeWare suite, as well as links
+  to discussions the development of new features in the BeeWare suite,
+  and ideas for new tools for the suite.
+
+* The original `BeeWare Users`_ and `BeeWare Developers`_ mailing
+  lists, have been shut down, in favor of the `getting help`_ page
+  mentioned above. However, the earlier messages and topics posted
+  prior to its shutdown remain. See the `announcement`_ for more
```
phildini: announcement doesn't seem to link to anything here?

Issue #181, opened 2016-07-21
Bytecode changes for dict in python 3.6 causes batavia.core.Dict to fail


slick666:
This is a continuation of #26 that was fixed in 3.4 and 3.5.. Python 3.6 fails for this same input but for different reasons. 

With python 3.5 the bytecode for storing data to the dict has changed to:

  2           0 LOAD_CONST               1 ('a')
              3 LOAD_CONST               2 (1)
              6 LOAD_CONST               3 ('b')
              9 LOAD_CONST               4 (2)
             12 BUILD_MAP                2
             15 STORE_FAST               0 (lonk)
causing dicts to be empty.

To reproduce use (in the testbed):

foo = {'a': 1, 'b': 2}
print('bar', foo)
Expected result: bar 1
Actual result: bar

cjrh:
Some exploration for 3.6:

```python
>>> code = '''\
... foo = {'a': 1, 'b': 2}
... print('bar', foo)'''
>>> code
"foo = {'a': 1, 'b': 2}\nprint('bar', foo)"
>>> cobj = compile(code, 'blah', 'exec')
>>> cobj
<code object <module> at 0x10889d420, file "blah", line 1>
>>> import dis
>>> dis.dis(cobj)
  1           0 LOAD_CONST               0 (1)
              2 LOAD_CONST               1 (2)
              4 LOAD_CONST               2 (('a', 'b'))
              6 BUILD_CONST_KEY_MAP      2
              8 STORE_NAME               0 (foo)

  2          10 LOAD_NAME                1 (print)
             12 LOAD_CONST               3 ('bar')
             14 LOAD_NAME                0 (foo)
             16 CALL_FUNCTION            2
             18 POP_TOP
             20 LOAD_CONST               4 (None)
             22 RETURN_VALUE
```

cjrh:
There are a bunch of new opcodes, so I'm adding these to `batavia/batavia/modules/dis.js`.  I'll also add (empty to start) implementations in `batavia/batavia/VirtualMachine.js`.

cjrh:
I've implemented the `BUILD_CONST_KEY_MAP` javasscript code (in the virtual machine) for Python 3.6 bytecode, in PR #536.  However, due to the wordcode change, there is still a lot of other things that are not working.

cjrh:
Note that in PR #536, this snippet:

```
foo = {'a': 1, 'b': 2}
print('bar', foo)
```

[...]

Issue #71, opened 2016-05-26
CPython 3.6+ is switching to a 16-bit wordcode format


freakboy3742:
Via @ncoghlan, reporting pybee/voc#132: 

This is an early heads up that CPython will be switching to a 16-bit wordcode format as of 3.6a2: http://bugs.python.org/issue26647

The change itself is in https://hg.python.org/cpython/rev/3a57eafd8401

In theory this should be a transparent change for any project that interacts with compiled code solely through the `dis` module, but since VoC transpiles CPython bytecode I figured you may want to take a closer look at the implications.

freakboy3742:
See also #26, for the changes affecting CPython 3.5.

dibyadas:
Now that VOC uses AST to do it, do you think this is issue is relevant any more?

cjrh:
I've begun implementation of 16-bit wordcode in PR #536. The changes are wound up in also trying to get dicts to work for 3.6.1. Unfortunately, both the 16-bit wordcode change as well as the _new_ opcode changes (like the new dict support) have to be done at the same time.

cjrh:
I've begun the process of using the 16-bit wordcode format in PR #536.

Zac-HD:
Closed by #621?

freakboy3742:
@Zac-HD Indeed it is. Thanks!

PR #536, opened 2017-05-23
Issue #181: python36 bytecode compatibility (for dicts and others)

cjrh:
This PR is still a WIP for discussion, since several details still need to be addressed:

- CPython 3.6 bytecodes are now 2 bytes instead of 1 byte (Issue #71)
- The implementations of the new opcodes are all still empty
- I have not yet looked at the specific issue in #181 (about dicts specifically)
- I have not added any test code yet. 

Note that this PR also includes implementations for the `__matmul__` and `__imatmul__` methods for the `BINARY_MATRIX_MULTIPLY` and `INPLACE_MATRIX_MULTIPLY` special methods. These changes are not required for 3.6 compat but they were easy to add.  They can be moved into a separate PR if desired.

Another subtlety: the `WITH_CLEANUP` bytecode name is now, at least in 3.6, separated into `WITH_CLEANUP_START` and `WITH_CLEANUP_END` and so I've changed these in `dis.js`. The Batavia code handling this opcode in the VM code is currently commented out, so I've gone ahead and changed `WITH_CLEANUP` into `WITH_CLEANUP_START` _inside_ the commented-out code. 

Before too much more work is done, we should discuss whether this is the right approach or not.

### DIFF
--- a/batavia/VirtualMachine.js
+++ b/batavia/VirtualMachine.js
@@ -289,6 +289,18 @@ VirtualMachine.prototype.build_dispatch_table = function() {
                             this.push(items[0] | items[1])
                         }
                     }
+                case 'MATRIX_MULTIPLY':
+                    return function() {
+                        var items = this.popn(2)
+                        if (items[0] === null) {
+                            this.push(types.NoneType.__matmul__(items[1]))
+                        } else if (items[0].__matmul__) {
+                            this.push(items[0].__matmul__(items[1]))
+                        } else {
+                            // TODO: This default action might be misleading.
+                            this.push(items[0] * items[1])
+                        }
+                    }
                 default:
                     throw new builtins.BataviaError.$pyclass('Unknown binary operator ' + operator_name)
             }
@@ -495,6 +507,24 @@ VirtualMachine.prototype.build_dispatch_table = function() {
                             }
                         } else {
                             items[0] |= items[1]
+                            result = items[0]
+                        }
+                        this.push(result)
+                    }
+                case 'MATRIX_MULTIPLY':
+                    return function() {
+                        var items = this.popn(2)
+                        var result
+                        if (items[0] === null) {
+                            result = types.NoneType.__imul__(items[1])
+                        } else if (items[0].__imul__) {
+                            result = items[0].__imul__(items[1])
+                            if (result === null) {
+                                result = items[0]
+                            }
+                        } else {
+                            // TODO: fallback multiply might be misleading
+                            items[0] *= items[1]
                             result = items[0]
                         }
                         this.push(result)
--- a/batavia/VirtualMachine.js
+++ b/batavia/VirtualMachine.js
@@ -1832,7 +1832,7 @@ VirtualMachine.prototype.byte_CALL_FUNCTION_KW = function(arg) {
     return this.call_function(arg, null, kwargs)
 }
 
-VirtualMachine.prototype.byte_CALL_FUNCTION_VAR_KW = function(arg) {
+VirtualMachine.prototype.byte_CALL_FUNCTION_EX = function(arg) {
     var items = this.popn(2)
     return this.call_function(arg, items[0], items[1])
 }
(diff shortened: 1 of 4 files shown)


cjrh:
So it turns out I was referring to a quite out-of-date branch of cpython.  I'll have to go through my updates all over again 🙃

cjrh:
Updated the opcodes to the correct list for CPython 3.6.  The next two actions are to (a) deal with 16bit opcodes, and (b) implement the new, empty, opcode handlers.

[This issue from b.p.o.](https://bugs.python.org/issue26647) explains in quite a lot of detail the rationale behind having 16-bit opcodes.

glasnt:
Hi @cjrh 

I'm running this against 3.6.1 in https://circleci.com/gh/pybee/batavia/834 (against the topic/cjrh-issue-181-bytecode-python36 branch). Let's see how this goes!

cjrh:
@glasnt Excellent, thanks!

cjrh:
For convenience, I'm duplicating the link to the 16-bit opcode change in CPython, [here](https://hg.python.org/cpython/rev/3a57eafd8401).

glasnt:
@cjrh there are a bunch of issues that the CI tests picked up. Take a look at the results and see what happened.

cjrh:
Doing a spot-check on this test: `./batavia/tests/builtins/test_abs.py`.  I'm using a debugger to examine the section in `tests.utils.TranspileTestCase#assertCodeExecution` where `py_out` is being compare to `js_out` (and I'm running this PR, under CPython 3.6.1).  

In the scenario I'm looking at, the *input* code is:

```
'try:
    print(\\'>>> f = abs\\')
    print(\\'>>> x = None\\')
    print(\\'>>> f(x)\\')
    f = abs
    x = None
    print(f(x))
except Exception as e:
    print(type(e), \\':\\', e)
print()
'
```

So, the **py_out** is

```
>>> f = abs
>>> x = None
>>> f(x)
<class 'TypeError'> : bad operand type for abs(): 'NoneType'
```

while the **js_out** is 

```
>>> f = abs
>>> x = None
>>> f(x)
### EXCEPTION ###
TypeError: bad operand type for abs(): \\'NoneType\\'
    test.py:7
```

Now I'm no batavia expert, but it _seems_ to me that the correct `TypeError` was raise in both, it's just that the text formatting is...different.  I guess this means we'll also need to change batavia's error reporting to match what cpython 3.6.1 produces, is that correct?

I expect such differences probably account for many of the failed tests in this PR right now.

cjrh:
The regex for doing `cleanse` on the python output will not match the output from 3.6.1, apparently.  That regex is at `tests/utils.py:172`, and is

```
PYTHON_EXCEPTION = re.compile('Traceback \(most recent call last\):\r?\n(  File "(?P<file>.*)", line (?P<line>\d+), in .*\r?\n    .*\r?\n)+(?P<exception>.*?): (?P<message>.*\r?\n)')  # NOQA
``` 

This will need to be changed too (for `cleanse` to work).

cjrh:
Following discussion with @freakboy3742, it seems that the **js_out** above is incorrect, since the exception should be getting handled and a traceback should not be printed.  This must be a regression introduced by the changes in this PR, so this must be investigated.

cjrh:
Testing this code snippet:

```
try:
    f = abs
    x = None
    print(f(x))
except Exception as e:
    print(type(e), ':', e)
```

Produces the correct output in the Batavia testbed in the browser.  Continuing to look for why the tests are different.

cjrh:
For interest sake, here is the side-by-side diff of the bytecode for py35 and py36 for the code in the previous comment:

image

cjrh:
I'm going to refactor the `unpack_code()` function in `batavia/VirtualMachine.js` so that the code for Python 3.5.x and Python 3.6.x are in separate functions.  The logic is too complex to deal with using these nested `if` statements testing for `constants.BATAVIA_MAGIC`, and also, deeply nested if's are a smell.

cjrh:
I may have successfully convinced @bspink to have a look at this too!

cjrh:
@glasnt Did @abonie complete 3.6 compat as part of SOC?  If so this PR should be closed.

freakboy3742:
@cjrh Yes, he did. There are still some test failures due to message format, but those exist on 3.5 as well; so I'm going close this ticket. Thanks for your work on this, even though it didn't end up being merged!

PR #539, opened 2017-05-24
Implement the sentinel variant of the 'iter' builtin

Nurdok:
The sentinel behavior is described here: https://docs.python.org/2/library/functions.html#iter


Nurdok:
@glasnt I added another test (and a fix).

Nurdok:
@glasnt Thanks!

PR #544, opened 2017-05-24
divmod() support for complex args and wrong number of args

njvrzm:


tests/builtins/test_divmod.py
```diff
@@ -8,27 +8,6 @@ class DivmodTests(TranspileTestCase):
 class BuiltinDivmodFunctionTests(BuiltinFunctionTestCase, TranspileTestCase):
     functions = ["divmod"]
 
-    not_implemented = [
```
phildini: All these tests just work? Huh... That's... suspicious? Maybe that's true! If so, wow!

njvrzm:
They didn't work until I fixed the exception string! But that seems to have been all they needed.

tests/builtins/test_divmod.py
freakboy3742: Makes sense, though - if the error message has been corrected, then most of these will have been failing because the message was wrong.

PR #548, opened 2017-05-25
adding builtin bytearray initializtion from integer

AndrewArcher:
Bytearrays can be now initialized with integers. In the Batavia testbed, running:
```
print(bytearray(2))
print(bytearray(-2))
```
outputs:
```
bytearray(b'\x00\x00')
  File "/var/folders/wk/ps21mr6j6ws9g_3lf8_81s75k_v486/T/tmpp_abue3j", line 3, in <module>
ValueError: negative count
```
just like python.

### DIFF
--- a/batavia/builtins/bytearray.js
+++ b/batavia/builtins/bytearray.js
@@ -1,3 +1,5 @@
+var Buffer = require('buffer').Buffer
+
 var exceptions = require('../core').exceptions
 var types = require('../types')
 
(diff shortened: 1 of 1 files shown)


glasnt:
Hi @AndrewArcher. Could you please add some tests, like your output suggests, to this PR?

robcarney:
A couple problems you'll encounter here: there's a specific hierarchy of exceptions being thrown here, if you were to expose the bytearray initialization to the testing framework you would get a few failures. For example, you need to throw an OverflowError for numbers where the bit size exceeds a certain threshold (2^63), a MemoryError for yet a smaller bit size threshold (2^31) (you would see these failures if you removed `test_int` from the not implemented section of the `BuiltinBytearrayFunctionTests` class). I also put in a pull request to fix this, but reverted the commit since it seems like there's another PR that solves this problem (along with other constructor overloads, see: pull request #550).

shaib:
... and this has already been implemented, as far as I can see.

PR #550, opened 2017-05-25
Implement bytearray on all argument types

ilankham:
All tests for builtin function bytearray now pass, with only small_ints tested and the convention that the behavior of bytearray(int) depends on the Javascript engine to determine whether an OverflowError occurs.

This is out of sync with 64-bit Python's behavior for large-magnitude numbers, which distinguishes between the following three cases:

- OverflowError(cannot fit 'int' into an index-sized integer) when int >= 2**63 or int < -2**63

- ValueError(negative count) when -2**63 <= int < 0

- MemoryError() when 2**31 < int < 2**63


shaib:
This seems to have already been implemented.

PR #553, opened 2017-05-26
divmod() support for incorrect argument types

njvrzm:


njvrzm:
This is follow-on work from #544, the commit for which unfortunately got in under my work email ^_^

PR #555, opened 2017-05-28
str.format

jstoebel:
when complete, this will implement `str.format`


wolfv:
Maybe it also makes sense to incorporate some tests from `brython`: https://github.com/brython-dev/brython/blob/master/www/tests/test_string_format.py

jstoebel:
hmmm, now there's a bunch of failing tests and I'm not sure what I did to fail them. Just out of curiosity I looked at the first one.

```
FAIL: test_abs_not_implemented (tests.builtins.test_abs.AbsTests)
  File "/Users/stoebelj/batavia/tests/builtins/test_abs.py", line 14, in test_abs_not_implemented
    """, run_in_function=False)
  File "/Users/stoebelj/batavia/tests/utils.py", line 446, in assertCodeExecution
AssertionError: '### EXCEPTION ###\nTypeError: TypeError:[48 chars]:6\n' != "### EXCEPTION ###\nTypeError: bad operan[43 chars]:6\n"
  ### EXCEPTION ###
- TypeError: TypeError: types.isinstance is not a function
+ TypeError: bad operand type for abs(): 'NotAbsLike'
      test.py:6
 : Global context
```

inside `abs.js` I moved `var types = require('../types')` inside of the `abs` function and the test passes.

Any ideas where I've buggered things up? I'm stumped.

jstoebel:
UPDATE: [This PR](https://github.com/pybee/batavia/pull/582) passes. it touches three files and it appears that the edits I made to those same files in _this_ PR are effectively the same. If anyone ha any ideas on where the problem might lie I'd love to know.

jstoebel:
I've rollback changes to `hash` and `types` and I'm still getting many errors. I'm stumped as to why.

tests/builtins/test_js2py.py
```diff
@@ -0,0 +1,8 @@
+from .. utils import TranspileTestCase
```
freakboy3742: This file is redundant; I think it's an artefact of an earlier merge with the j2py patch.

jstoebel: file removed!

PR #568, opened 2017-06-12
Fix List class index method and add unittest

chairco:


tests/datatypes/test_list.py
```diff
@@ -124,6 +124,35 @@ def test_getitem(self):
             print(x[-10])
             """)
 
+    def test_index(self):
+        self.assertCodeExecution("""
+        x = (1, 2, 2, 3)
```
freakboy3742: x here will be a tuple, not a list; is this what you intended?

chairco: Yes, is wrong type. and fix it. by the way, I push wrong code |0|.

chairco:
Python's list.index(x[, start[, end]]) allow negative numbers. when the parameter is negative do some convert.

example:
```
a = [2,3,4,5]
print(a.index(2, -4, -4))
```
convert the start value: len(a) + (-4) = 0, but after caculate still negative, assign  start value = 0

PR #572, opened 2017-06-13
A more complete ord builtin function

adrianliaw:
- Add argument checking in `ord` implementation
- Only accept `str`, `bytes` or `bytearray` with length 1
- Returns a `types.Int` object
- Add tests for `ord`

Additional patches:
- Return `types.Int` in `str.__len__` method
- Fix `bytes.__getitem__` method

Known problems:
- Unicode characters whose UTF-16 encoding is greater than 16 bits (e.g. emojis) are failed to handle

### DIFF
--- a/batavia/builtins/ord.js
+++ b/batavia/builtins/ord.js
@@ -5,27 +5,28 @@ var types = require('../types')
 
 function ord(args, kwargs) {
     if (arguments.length !== 2) {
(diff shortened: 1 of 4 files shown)


freakboy3742:
Good news - the test failures are all linting problems. If you run eslint over the batavia directory (or just correct the problems as listed in the error trace) you'll be ready for a merge!

adrianliaw:
@freakboy3742 Updated!

PR #573, opened 2017-06-17
This commit fixes #367

### LINKED ISSUE #367: Some builtins should raise `TypeError: foo() takes no keyword arguments`
A number of builtin methods implemented in C in CPython can never take keyword arguments; examples include `tuple.count()` and `tuple.index`. If you try to pass any keyword argument, you'll get `TypeError: xxx() takes no keyword arguments`. This should be reproduced in batavia.

krtk6160:
Fixed keyword argument handling for `tuple.count()` and `tuple.index()`

### DIFF
--- a/batavia/types/Tuple.js
+++ b/batavia/types/Tuple.js
@@ -534,11 +534,15 @@ Tuple.prototype.copy = function() {
     return new Tuple(this)
 }
 
-Tuple.prototype.count = function(value) {
-    if (arguments.length !== 1) {
-        throw new exceptions.TypeError.$pyclass('count() takes exactly one argument (' + arguments.length + ' given)')
+Tuple.prototype.count = function(args, kwargs) {
+    if (args.length !== 1) {
+        throw new exceptions.TypeError.$pyclass('count() takes exactly one argument (' + args.length + ' given)')
+    }
+    if (kwargs && Object.keys(kwargs).length > 0){
+        throw new exceptions.TypeError.$pyclass('count() takes no keyword arguments')
     }
     var count = 0
+    var value = args[0]
     for (var i = 0; i < this.length; ++i) {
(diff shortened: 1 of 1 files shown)


batavia/types/Tuple.js
```diff
@@ -547,12 +551,20 @@ Tuple.prototype.count = function(value) {
...
+    if (args.length < 1) {
+        throw new exceptions.TypeError.$pyclass('index() takes at least 1 argument (' + args.length + ' given)')
+    } else if (args.length > 3) {
+        throw new exceptions.TypeError.$pyclass('index() takes at most 3 arguments (' + args.length + ' given)')
     }
+    if (kwargs && Object.keys(kwargs).length > 0) {
+        throw new exceptions.TypeError.$pyclass('index() takes no keyword arguments')
+    }
+    var value = args[0]
+    if (args[1]) { var start = args[1] }
```
freakboy3742: The linter apparently doesn't catch this, but single-line if statements aren't preferred style.

batavia/types/Tuple.js
```diff
@@ -547,12 +551,20 @@ Tuple.prototype.count = function(value) {
     return count
 }
 
-Tuple.prototype.index = function(value, start, stop) {
-    if (arguments.length < 1) {
-        throw new exceptions.TypeError.$pyclass('index() takes at least 1 argument (' + arguments.length + ' given)')
-    } else if (arguments.length > 3) {
-        throw new exceptions.TypeError.$pyclass('index() takes at most 3 arguments (' + arguments.length + ' given)')
+Tuple.prototype.count.$pyargs = true
```
freakboy3742: I can see what you're doing here, but I'm not convinced it's the right approach. I think the fix is a little higher up, in the dispatch mechanism itself.

PR #582, opened 2017-06-27
fix js2py for object -> dict

jstoebel:
- require `types` inside of hash function
 - use `__setitem__` in js2py rather than tacking on a new attr

### DIFF
--- a/batavia/builtins/hash.js
+++ b/batavia/builtins/hash.js
@@ -1,8 +1,8 @@
 var exceptions = require('../core').exceptions
 var type_name = require('../core').type_name
-var types = require('../types')
 
 function hash(args, kwargs) {
+    var types = require('../types')
     if (arguments.length !== 2) {
         throw new exceptions.BataviaError.$pyclass('Batavia calling convention not used.')
     }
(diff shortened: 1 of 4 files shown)


batavia/builtins/hash.js
```diff
@@ -1,8 +1,8 @@
 var exceptions = require('../core').exceptions
 var type_name = require('../core').type_name
-var types = require('../types')
 
 function hash(args, kwargs) {
+    var types = require('../types')
```
dimaqq: Is this change intended?

Node.js advises against `require` inside function body:
https://github.com/alanjames1987/Node.js-Best-Practices/issues/3

freakboy3742: Node.js may well recommend that - but have they provided an alternative way to resolve circular dependencies in imports?

dimaqq: It's a fair point that something must be done.
However require on each call to `hash()` seems like a crazy penalty, doesn't it?

I'm assuming it's going to be used a lot.
P.S. I don't see `hash` being used by `types` anyway...

jstoebel: @dimaqq I moved `var types = require('../types')` because I was otherwise unable to access `types.isinstance` inside of the `hash` function

jstoebel:
@freakboy3742 yes, I will put together some tests for `js2py`

PR #586, opened 2017-07-04
Json decoding

abonie:
Basic features of Python's `JSONDecoder` implemented with JavaScript's `JSON.parse`.


freakboy3742:
Looks like there's a merge required here as a result of the test_json naming change. If you can sort that out, I'll push the button.

abonie:
I think it is ready, but CI did not rerun tests.

PR #587, opened 2017-07-05
Changed directory during Windows venv setup (docs)

AlexLloyd1:
After 'activate'-ing the venv, you're still in ...\pybee\venv\Scripts.
Need to move back to ...\batavia in order to pip install -e . (otherwise you get 'Directory '.' is not installable. File 'setup.py' not found.')


freakboy3742:
Good catch. I've tweaked the instructions slightly to avoid bouncing around directories, and to maintain consistency with the Mac/Linux instructions.

PR #588, opened 2017-07-07
fix generated testnames for json module not valid python identifiers

### LINKED ISSUE #585: cricket-unittest fails to load tests
Attempting to run `cricket-unittest` in the project's root directory, as per the docs, bails immediately with the following:

```
  File "/home/jms/.local/share/snakefarm/batavia-34/bin/cricket-unittest", line 11, in <module>
    sys.exit(main())
  File "/home/jms/.local/share/snakefarm/batavia-34/lib/python3.4/site-packages/cricket/unittest/__main__.py", line 9, in main
    cricket_main(UnittestProject)
  File "/home/jms/.local/share/snakefarm/batavia-34/lib/python3.4/site-packages/cricket/main.py", line 74, in main
    project.refresh(test_list, errors)
  File "/home/jms/.local/share/snakefarm/batavia-34/lib/python3.4/site-packages/cricket/model.py", line 506, in refresh
    testModule._purge(timestamp)
  File "/home/jms/.local/share/snakefarm/batavia-34/lib/python3.4/site-packages/cricket/model.py", line 381, in _purge
    testModule._purge(timestamp)
  File "/home/jms/.local/share/snakefarm/batavia-34/lib/python3.4/site-packages/cricket/model.py", line 381, in _purge
    testModule._purge(timestamp)
  File "/home/jms/.local/share/snakefarm/batavia-34/lib/python3.4/site-packages/cricket/model.py", line 381, in _purge
    testModule._purge(timestamp)
  File "/home/jms/.local/share/snakefarm/batavia-34/lib/python3.4/site-packages/cricket/model.py", line 242, in _purge
    if testMethod.timestamp != timestamp:
AttributeError: 'TestCase' object has no attribute 'timestamp'
``

This looks to me like it could be a problem upstream with `cricket`, given that no `batavia` specific files are mentioned in the tb, but it could also have to do with how `batavia`'s tests are formatted? I unfortunately don't know either project well enough to determine which project gets the bug report, so I'm submitting it here.

For what its worth, I'm on Arch Linux, and in three separate virtual environments (Python 3.4.6, 3.5.3, and 3.6.1) using cricket 0.2.5, I get the same traceback. IDLE and tk are properly installed. Running the tests via vanilla unittest or pytest both behave as expected.  If I descend into the `batavia` build directory after running `npm run build`, `cricket-unittest` opens without error but (unsurprisingly) fails to load any tests.

jmsdvl:
Some of the tests generated in `tests/modules/test_json.py` produced test cases with names that are not valid Python identifiers; this breaks tools such as `cricket` which attempt to load and manipulate tests based on import paths.  This fixes #585 and is relevant to pybee/cricket#39.

### DIFF
--- a/tests/modules/test_json.py
+++ b/tests/modules/test_json.py
@@ -48,12 +48,12 @@ class JSONEncoderTests(ModuleFunctionTestCase, TranspileTestCase):
         """)
 
     not_implemented = [
-        'test_json_JSONEncoder().encode_dict',   # fails due to dict ordering
-        'test_json_JSONEncoder().encode_class',  # fails due to class __str__
+        'test_json_JSONEncoder__encode_dict',   # fails due to dict ordering
+        'test_json_JSONEncoder__encode_class',  # fails due to class __str__
     ]
 
 
-JSONEncoderTests.add_one_arg_tests('json', ['JSONEncoder().encode'])
+JSONEncoderTests.add_one_arg_tests('json', ['JSONEncoder__encode'])
 
 
 class DumpsTests(ModuleFunctionTestCase, TranspileTestCase):


robcarney:
I was having the same issue, and the change in the commit here fixed it for me.

jmsdvl:
You're right, all the tests that were named `JSONEncoder().<method>` now fail.  It looks like at some point the testing apparatus is _executing_ the test name, or more accurately (which would have prevented me making this pull request if it had occurred to me), the test method is being named the same thing as a string its also supposed to evaluate.

abonie:
Can you check for me if this https://github.com/pybee/batavia/pull/589 fixes your issue? I don't have cricket set up (that's why this bug slipped in in the first place) and don't have much time right now, but I slapped this fix together and I think it should solve the problem.

PR #589, opened 2017-07-08
Fix issue with cricket and invalid test names

abonie:
An attempt to fix issue described here: https://github.com/pybee/batavia/pull/588

### DIFF
--- a/tests/modules/test_json.py
+++ b/tests/modules/test_json.py
@@ -1,8 +1,9 @@
 from itertools import product
-from ..utils import TranspileTestCase, ModuleFunctionTestCase, adjust
+from ..utils import TranspileTestCase, ModuleFunctionTestCase, MethodTestCase, \
+    adjust
 
 
-class JSONEncoderTests(ModuleFunctionTestCase, TranspileTestCase):
+class JSONEncoderTests(MethodTestCase, TranspileTestCase):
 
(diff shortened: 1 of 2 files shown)


jmsdvl:
This fixes the errors with cricket trying to load the JSONEncoder tests, but now all the tests in `StdlibTests` are failing from `tests/utils.py:445` with some variant on:

```
AssertionError: 'Unknown opcode 0 (<0>)\n' != 'Done\n'
- Unknown opcode 0 (<0>)
+ Done
```

Where done is just the expected output for this test (`test_operator` in this case).

abonie:
@jmsdvl Are you sure no other changes affected this outcome? It's either that, or some issue with cricket I have no idea how to deal with right now. Could you try and run tests from command line e.g. `python setupy.py test -s tests/modules/test_stdlib.py`? I get no failures then, and also I don't see how these changes could affect StdlibTests at all.

jmsdvl:
>I don't see how these changes could affect StdlibTests at all

@abonie I don't see how either.  I re-ran the tests in a new, clean environment this morning and all tests pass.  I have no idea what was causing the failures - no other source code was changed (according to `git` at least), and I was seeing the failures both from cricket and `python setup.py test`.  But not now ಠ_ಠ 

Looks like this is a good patch then.

PR #594, opened 2017-07-31
`with` statement

abonie:
Implement SETUP_WITH and WITH_CLEANUP opcodes in Batavia's VM

### DIFF
--- a/batavia/VirtualMachine.js
+++ b/batavia/VirtualMachine.js
@@ -619,6 +619,15 @@ VirtualMachine.prototype.push = function(val) {
 }
 
 /*
+ * Push value onto the stack, i elements behind TOS
+ * push_at(val, 0) is equivalent to push(val)
+ * push_at(val, 1) will result in val being second on the stack
+ */
+VirtualMachine.prototype.push_at = function(val, i) {
+    this.frame.stack.splice(this.frame.stack.length - i, 0, val)
+}
+
+/*
  * Pop a number of values from the value stack.
  *
  * A list of `n` values is returned, the deepest value first.
@@ -646,7 +655,7 @@ VirtualMachine.prototype.jump = function(jump) {
 }
 
 VirtualMachine.prototype.push_block = function(type, handler, level) {
-    if (level === null) {
+    if (level === null || level === undefined) {
         level = this.frame.stack.length
     }
     this.frame.block_stack.push(new Block(type, handler, level))
@@ -1638,6 +1647,10 @@ VirtualMachine.prototype.byte_END_FINALLY = function() {
     var exc_type = this.pop()
     if (exc_type === builtins.None) {
         why = null
+    } else if (exc_type === 'silenced') {
+        var block = this.pop_block() // should be except-handler
+        this.unwind_block(block)
+        return null
     } else {
         value = this.pop()
         if (value instanceof builtins.BaseException.$pyclass) {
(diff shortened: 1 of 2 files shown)


batavia/VirtualMachine.js
```diff
@@ -1716,16 +1720,43 @@ VirtualMachine.prototype.byte_POP_EXCEPT = function() {
...
+    } else if (exc.$pyclass.prototype instanceof exceptions.BaseException.$pyclass) {
+        val = this.peek(1)
+        tb = this.peek(2)
+        exit_func = this.pop(6)
+        this.frame.stack.splice(-3, 0, builtins.None)  // XXX
+        var block = this.pop_block()
+        this.push_block(block.type, block.handler, block.level-1)
+    } else {
+        throw new builtins.BataviaError.$pyclass('Confused WITH_CLEANUP')
+    }
```
brutusthebee: At column 56: [(no-multi-spaces) Multiple spaces found before '// XXX'.](http://eslint.org/docs/rules/no-multi-spaces)

batavia/VirtualMachine.js
```diff
@@ -1716,16 +1720,43 @@ VirtualMachine.prototype.byte_POP_EXCEPT = function() {
...
+    var ret = callables.call_function(exit_func, [exc, val, tb]) // XXX this arg list is incorrect for first 2 cases
+    //TODO: silence error if necessary
```
brutusthebee: At column 63: [(space-infix-ops) Infix operators must be spaced.](http://eslint.org/docs/rules/space-infix-ops)

batavia/VirtualMachine.js
```diff
@@ -1716,16 +1720,43 @@ VirtualMachine.prototype.byte_POP_EXCEPT = function() {
...
+    if (!(exc instanceof types.NoneType) && ret) { // XXX
+        this.push('silenced')
+    }
+}
+
```
brutusthebee: At column 5: [(spaced-comment) Expected space or tab after '//' in comment.](http://eslint.org/docs/rules/spaced-comment)

batavia/VirtualMachine.js
```diff
@@ -1716,16 +1720,43 @@ VirtualMachine.prototype.byte_POP_EXCEPT = function() {
...
+    } else if (exc.$pyclass.prototype instanceof exceptions.BaseException.$pyclass) {
+        val = this.peek(1)
+        tb = this.peek(2)
+        exit_func = this.pop(6)
+        this.frame.stack.splice(-3, 0, builtins.None)  // XXX
+        var block = this.pop_block()
+        this.push_block(block.type, block.handler, block.level - 1)
+    } else {
+        throw new builtins.BataviaError.$pyclass('Confused WITH_CLEANUP')
+    }
```
brutusthebee: At column 56: [(no-multi-spaces) Multiple spaces found before '// XXX'.](http://eslint.org/docs/rules/no-multi-spaces)

brutusthebee: At column 56: [(no-multi-spaces) Multiple spaces found before '// XXX'.](http://eslint.org/docs/rules/no-multi-spaces)

freakboy3742: Not sure what's going on here - the lint problem is genuine; but it's on line 1747.

abonie: I fixed it. The `XXX` was only there to remind me to double check that this line works as intended. But now I am wondering if I should leave it as is, since it breaks "stack API" by using `this.frame.stack.slice` instead of combination of `this.pop` and `this.push`. The other way feels clunky in this case.

batavia/VirtualMachine.js
```diff
@@ -1716,16 +1720,43 @@ VirtualMachine.prototype.byte_POP_EXCEPT = function() {
...
-//         this.push(ctxmgr.__exit__)
-//         ctxmgr_obj = ctxmgr.__enter__()
-//         if PY2:
-//             this.push_block('with', dest)
-//         elif PY3:
-//             this.push_block('finally', dest)
-//         this.push(ctxmgr_obj)
-// }
+VirtualMachine.prototype.byte_SETUP_WITH = function(dest) {
+    var mgr = this.top()
```
abonie: CPython would pop context manager from the stack here and push only `__exit__` method of the context manager to call it  as a function in `WITH_CLEANUP`. However I couldn't make it work this way, so I am leaving context manager on the stack and in `WITH_CLEANUP` I am calling `mgr.__exit__` (line 1753). I don't see how this would break anything. What do you think @freakboy3742 ?

freakboy3742: Agreed. The state of the stack is an internal detail; and in this particular case, it's a transient detail - so go right ahead.

PR #607, opened 2017-08-07
Update existing PowTests

qilyn:
This PR should remove the need for `@expectedFailure` for the six existing tests in `tests/builtins/test_pow.py` by raising the correct Python exception. 

The exception message for a `pow()` with non-integer looks like it might be valid for another version of Python? On v3.5.2, this change fixes it.


shaib:
This seems to be implememted more correctly by #746 (it also appears to include a redundant, old version of `batavia/builtins/pow.js` as `builtins/pow.js`)

PR #611, opened 2017-08-07
Allows calling 'set' on a 'dict'

filipeximenes:


filipeximenes:
No longer needed. Covered in #613

PR #616, opened 2017-08-08
Updated pow error message

Lewis-Quill:
Updated the error message.
Got 4 unexpected successes in cricket!


Lewis-Quill:
I'm guessing I should have updated test_pow.py ??

shaib:
I believe this is obsolete now

PR #620, opened 2017-08-08
Split WITH_CLEANUP into WITH_CLEANUP_START/FINISH

abonie:
Opcode 81 - WITH_CLEANUP is replaced in Python 3.5+ by pair of opcodes:
81 - WITH_CLEANUP_START and 82 - WITH_CLEANUP_FINISH

### DIFF
--- a/batavia/VirtualMachine.js
+++ b/batavia/VirtualMachine.js
@@ -1761,6 +1761,27 @@ VirtualMachine.prototype.byte_WITH_CLEANUP = function() {
         throw new builtins.BataviaError.$pyclass('Confused WITH_CLEANUP')
     }
     var ret = callables.call_method(mgr, '__exit__', [exc, val, tb])
+    if (constants.BATAVIA_MAGIC === constants.BATAVIA_MAGIC_34) {
+        if (!(exc instanceof types.NoneType) && ret.__bool__ !== undefined &&
+                ret.__bool__().valueOf()) {
+            this.push('silenced')
+        }
+    } else {
+        // Assuming Python 3.5
+        this.push(exc)
+        this.push(ret)
+    }
+}
+
+VirtualMachine.prototype.byte_WITH_CLEANUP_FINISH = function() {
+    if (constants.BATAVIA_MAGIC === constants.BATAVIA_MAGIC_34) {
+        throw new builtins.BataviaError.$pyclass(
+            'Unknown opcode WITH_CLEANUP_FINISH in Python 3.4'
+        )
+    }
+    // Assuming Python 3.5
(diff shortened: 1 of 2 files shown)


batavia/VirtualMachine.js
```diff
@@ -1761,6 +1761,27 @@ VirtualMachine.prototype.byte_WITH_CLEANUP = function() {
...
+
+VirtualMachine.prototype.byte_WITH_CLEANUP_FINISH = function() {
+    if (constants.BATAVIA_MAGIC === constants.BATAVIA_MAGIC_34) {
+        throw new builtins.BataviaError.$pyclass(
+            'Unknown opcode ' + opcode + ' (' + opname + ')'
+        )
+    }
+    // Assuming Python 3.5
+    var ret = this.pop()
+    var exc = this.pop()
```
brutusthebee: At column 33: [(no-undef) 'opcode' is not defined.](http://eslint.org/docs/rules/no-undef)

brutusthebee: At column 49: [(no-undef) 'opname' is not defined.](http://eslint.org/docs/rules/no-undef)

phildini: I don't see how your code is failing this check, so I think this is safe to ignore while we still work out bugs in beekeeper.

phildini: See above

batavia/VirtualMachine.js
```diff
@@ -1761,6 +1761,27 @@ VirtualMachine.prototype.byte_WITH_CLEANUP = function() {
         throw new builtins.BataviaError.$pyclass('Confused WITH_CLEANUP')
     }
     var ret = callables.call_method(mgr, '__exit__', [exc, val, tb])
+    if (constants.BATAVIA_MAGIC === constants.BATAVIA_MAGIC_34) {
+        if (!(exc instanceof types.NoneType) && ret.__bool__ !== undefined &&
+                ret.__bool__().valueOf()) {
+            this.push('silenced')
+        }
+    } else {
+        // Assuming Python 3.5
```
phildini: Will this code work for 3.5+ or just 3.5?

freakboy3742: Depends on when/if they change the opcode. It looks like it's the same for 3.6 (except for the format change - but that's a marshalling issue, not an opcode issue.)

batavia/VirtualMachine.js
freakboy3742: Hrm... Yeah - not sure what's going on here, because the Beefore check passed...

PR #621, opened 2017-08-08
Parse wordcodes in Python 3.6

abonie:
Python 3.6 uses 2 bytes for each instruction (instead of varying number of bytes as in previous versions). This PR will introduce changes to VM that check for Python version to parse bytecode appropriately.

Single wordcode is 2 bytes wide and consists of an opcode and its
argument (or 0 if no argument is used)

TODO opcodes:
EXTENDED_ARG
???

### DIFF
--- a/batavia/VirtualMachine.js
+++ b/batavia/VirtualMachine.js
@@ -837,30 +837,9 @@ VirtualMachine.prototype.unpack_code = function(code) {
                 lo = code.co_code.val[pos++]
                 hi = code.co_code.val[pos++]
                 extra = (lo << 16) | (hi << 24)
-                // emulate four NOPs
+                // emulate NOP
                 unpacked_code[opcode_start_pos] = {
                     'opoffset': opcode_start_pos,
-                    'opcode': dis.NOP,
-                    'op_method': this.dispatch_table[dis.NOP],
-                    'args': [],
-                    'next_pos': pos
-                }
-                unpacked_code[opcode_start_pos + 1] = {
-                    'opoffset': opcode_start_pos + 1,
-                    'opcode': dis.NOP,
-                    'op_method': this.dispatch_table[dis.NOP],
-                    'args': [],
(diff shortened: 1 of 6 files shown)


cjrh:
Just a heads-up I've been tinkering with this in #536.  I didn't get very far, but I did get the new dict opcode change done.  (It's most other things that are still broken in my PR).

cjrh:
By all means please feel free to take anything that might be useful from my PR.  I have very little time available to work on this.

abonie:
Thanks cjrh, I have not touched the dict changes yet, so it works out.
For now I got sidetracked by the fact, that argument unpacking operators `*` and `**` do not work under any Python version, but I will definitely checkout your PR when I get to dicts.

abonie:
@cjrh if I were to use some of your code from #536 what is the proper way to give you credit? Should I just mention it in commit message?

cjrh:
@abonie Sorry for the delay!  No credit necessary, just go for it 👍

batavia/VirtualMachine.js
```diff
@@ -1428,6 +1495,18 @@ VirtualMachine.prototype.byte_BUILD_MAP = function(size) {
...
+    for (var i = 0; i < values.length; i += 1) {
+        dict.__setitem__(keys[i], values[i])
+    }
+    this.push(dict)
+    return
+}
+
 VirtualMachine.prototype.byte_STORE_MAP = function() {
     switch (constants.BATAVIA_MAGIC) {
         case constants.BATAVIA_MAGIC_35:
```
brutusthebee: At column 5: [(no-useless-return) Unnecessary return statement.](http://eslint.org/docs/rules/no-useless-return)

brutusthebee: At column 5: [(no-useless-return) Unnecessary return statement.](http://eslint.org/docs/rules/no-useless-return)

PR #636, opened 2017-08-17
Update contributing.rst

deatonjm:
added the line to upgrade npm to newest version

I added a line to upgrade npm in the "Contributing to Batavia" walk-through. It is present in the Tutorial, but if you are just using the "Contributing to Batavia" instructions, you miss that step and everything explodes.

Makes setting up your environment to begin contributing less confusing.

N/A

## PR Checklist:

- [? ] All new features have been tested
- [x ] All new features have been documented
- [x ] I have read the **CONTRIBUTING.md** file
- [x ] I will abide by the code of conduct

### DIFF
--- a/docs/community/contributing.rst
+++ b/docs/community/contributing.rst
@@ -57,8 +57,8 @@ install Batavia's JavaScript dependencies, and compile the Batavia library:
 
 .. code-block:: bash
 
-    $ npm install
     $ npm install -g npm
+    $ npm install
     $ npm run build
 
 .. _Node.js: https://nodejs.org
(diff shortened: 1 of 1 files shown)


docs/community/contributing.rst
```diff
@@ -58,6 +58,7 @@ install Batavia's JavaScript dependencies, and compile the Batavia library:
 .. code-block:: bash
 
     $ npm install
+    $ npm install -g npm
```
freakboy3742: Good idea - but this line needs to be run before `npm install`; if you do it this way, you'll end up with the install being done with the stale version of npm.

If you reverse the order of line 60 and 61, it should be good to go.

deatonjm: Will do, I thought you had told me to do them in this order explicitly, but I may have mis-remembered.

PR #643, opened 2017-08-20
Version dependent error messages

abonie:
There are multiple discrepancies in error messages for Python 3.5+, which causes 3.5 and 3.6 test suites to fail. This tries to fix them.

### DIFF
--- a/batavia/builtins/bytes.js
+++ b/batavia/builtins/bytes.js
@@ -3,6 +3,7 @@ var Buffer = require('buffer').Buffer
 var exceptions = require('../core').exceptions
 var callables = require('../core').callables
 var type_name = require('../core').type_name
+var constants = require('../core').constants
 var types = require('../types')
 var iter = require('./iter')
 
@@ -26,7 +27,19 @@ function bytes(args, kwargs) {
     } else if (args.length === 1) {
         var arg = args[0]
         if (arg === null) {
-            throw new exceptions.TypeError.$pyclass("'NoneType' object is not iterable")
+            switch (constants.BATAVIA_MAGIC) {
+                case constants.BATAVIA_MAGIC_34:
+                case constants.BATAVIA_MAGIC_35a0:
+                case constants.BATAVIA_MAGIC_35:
+                case constants.BATAVIA_MAGIC_353:
+                    throw new exceptions.TypeError.$pyclass(
+                        "'NoneType' object is not iterable"
+                    )
+                case constants.BATAVIA_MAGIC_36:
+                    throw new exceptions.TypeError.$pyclass(
+                        "cannot convert 'NoneType' object to bytes"
+                    )
+            }
         } else if (types.isinstance(arg, types.Int)) {
             // bytes(int) -> bytes array of size given by the parameter initialized with null bytes
             // Batavia ints are BigNumbers, so we need to unpack the value from the BigNumber Array.
@@ -98,9 +111,19 @@ function bytes(args, kwargs) {
             return new types.Bytes(Buffer.from(buffer_args))
         } else {
             // the argument is not one of the special cases, and not an iterable, so...
-            throw new exceptions.TypeError.$pyclass(
-            //    "'" + type_name(val) + "' object is not iterable");
-            "'" + type_name(arg) + "' object is not iterable")
+            switch (constants.BATAVIA_MAGIC) {
+                case constants.BATAVIA_MAGIC_34:
+                case constants.BATAVIA_MAGIC_35a0:
+                case constants.BATAVIA_MAGIC_35:
+                case constants.BATAVIA_MAGIC_353:
+                    throw new exceptions.TypeError.$pyclass(
+                        "'" + type_name(arg) + "' object is not iterable"
+                    )
+                case constants.BATAVIA_MAGIC_36:
+                    throw new exceptions.TypeError.$pyclass(
+                        "cannot convert '" + type_name(arg) + "' object to bytes"
+                    )
+            }
         }
     } else if (args.length >= 2 && args.length <= 3) {
         //    bytes(string, encoding[, errors]) -> bytes
--- a/batavia/builtins/chr.js
+++ b/batavia/builtins/chr.js
@@ -17,6 +17,7 @@ function chr(args, kwargs) {
             case constants.BATAVIA_MAGIC_35a0:
             case constants.BATAVIA_MAGIC_35:
             case constants.BATAVIA_MAGIC_353:
+            case constants.BATAVIA_MAGIC_36:
                 throw new exceptions.TypeError.$pyclass('chr() takes exactly one argument (' + args.length + ' given)')
 
             default:
--- a/batavia/core/exceptions.js
+++ b/batavia/core/exceptions.js
@@ -441,4 +441,13 @@ ZeroDivisionError.prototype.__class__.$pyclass = ZeroDivisionError
 
 exceptions.ZeroDivisionError = ZeroDivisionError.prototype.__class__
 
+var JSONDecodeError = function(msg) {
+    Exception.call(this, 'JSONDecodeError', msg)
+}
+JSONDecodeError.prototype = Object.create(Exception.prototype)
+JSONDecodeError.prototype.__class__ = new Type('JSONDecodeError', [Exception.prototype.__class__])
+JSONDecodeError.prototype.__class__.$pyclass = JSONDecodeError
+
+exceptions.JSONDecodeError = JSONDecodeError.prototype.__class__
+
 module.exports = exceptions
--- a/batavia/core/types/NoneType.js
+++ b/batavia/core/types/NoneType.js
@@ -49,7 +49,7 @@ NoneType.prototype.__setattr__ = function(attr, value) {
  **************************************************/
 
 NoneType.prototype.__lt__ = function(other) {
-    switch(constants.BATAVIA_MAGIC) {
+    switch (constants.BATAVIA_MAGIC) {
         case constants.BATAVIA_MAGIC_34:
         case constants.BATAVIA_MAGIC_35a0:
         case constants.BATAVIA_MAGIC_35:
@@ -66,7 +66,7 @@ NoneType.prototype.__lt__ = function(other) {
 }
 
 NoneType.prototype.__le__ = function(other) {
-    switch(constants.BATAVIA_MAGIC) {
+    switch (constants.BATAVIA_MAGIC) {
         case constants.BATAVIA_MAGIC_34:
         case constants.BATAVIA_MAGIC_35a0:
         case constants.BATAVIA_MAGIC_35:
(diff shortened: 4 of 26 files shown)


abonie:
Anyone knows how to handle situation where for certain test cases there are `expected failures` in Python 3.4 but `unexpected successes` for Python 3.5?

lucantrop:
How many of those instances are there? I only see 20 failures in the output from BeeKeeper (no unexpected successes).
Anyway, you might want to make a special decorator that handles those cases. It's only a matter of adding a version check to [`expectedFailure`](https://github.com/python/cpython/blob/master/Lib/unittest/case.py#L119).

batavia/VirtualMachine.js
```diff
@@ -755,95 +755,161 @@ VirtualMachine.prototype.create_traceback = function() {
...
+        code.co_unpacked_code = unpacked_code
+    } else {
+        // Until 3.6 Python had variable width opcodes
+
+        var pos = 0
+        var unpacked_code = []
+        var args
+        var extra = 0
+        var lo
+        var hi
```
brutusthebee: At column 13: [(no-redeclare) 'pos' is already defined.](http://eslint.org/docs/rules/no-redeclare)

batavia/VirtualMachine.js
```diff
@@ -755,95 +755,161 @@ VirtualMachine.prototype.create_traceback = function() {
...
+
```
brutusthebee: At column 13: [(no-redeclare) 'unpacked_code' is already defined.](http://eslint.org/docs/rules/no-redeclare)

batavia/VirtualMachine.js
```diff
@@ -755,95 +755,161 @@ VirtualMachine.prototype.create_traceback = function() {
...
+        while (pos < code.co_code.val.length) {
```
brutusthebee: At column 13: [(no-redeclare) 'args' is already defined.](http://eslint.org/docs/rules/no-redeclare)

batavia/VirtualMachine.js
```diff
@@ -755,95 +755,161 @@ VirtualMachine.prototype.create_traceback = function() {
...
+            var opcode_start_pos = pos
```
brutusthebee: At column 13: [(no-redeclare) 'extra' is already defined.](http://eslint.org/docs/rules/no-redeclare)

batavia/VirtualMachine.js
```diff
@@ -755,95 +755,161 @@ VirtualMachine.prototype.create_traceback = function() {
...
+
+            var opcode = code.co_code.val[pos++]
+
+            // next opcode has 4-byte argument effectively.
+            if (opcode === dis.EXTENDED_ARG) {
```
brutusthebee: At column 17: [(no-redeclare) 'opcode_start_pos' is already defined.](http://eslint.org/docs/rules/no-redeclare)

batavia/VirtualMachine.js
```diff
@@ -755,95 +755,161 @@ VirtualMachine.prototype.create_traceback = function() {
...
+                lo = code.co_code.val[pos++]
+                hi = code.co_code.val[pos++]
```
brutusthebee: At column 17: [(no-redeclare) 'opcode' is already defined.](http://eslint.org/docs/rules/no-redeclare)

batavia/VirtualMachine.js
```diff
@@ -755,95 +755,161 @@ VirtualMachine.prototype.create_traceback = function() {
...
             } else {
-                args = [intArg]
+                lo = code.co_code.val[pos++]
+                hi = code.co_code.val[pos++]
+                var intArg = lo | (hi << 8) | extra
+                extra = 0 // use extended arg if present
+
+                if (opcode in dis.hasconst) {
+                    args = [code.co_consts[intArg]]
+                } else if (opcode in dis.hasfree) {
```
brutusthebee: At column 21: [(no-redeclare) 'intArg' is already defined.](http://eslint.org/docs/rules/no-redeclare)

batavia/VirtualMachine.js
```diff
@@ -755,95 +755,161 @@ VirtualMachine.prototype.create_traceback = function() {
...
+                    if (intArg < code.co_cellvars.length) {
+                        args = [code.co_cellvars[intArg]]
+                    } else {
+                        var var_idx = intArg - code.co_cellvars.length
+                        args = [code.co_freevars[var_idx]]
+                    }
+                } else if (opcode in dis.hasname) {
+                    args = [code.co_names[intArg]]
+                } else if (opcode in dis.hasjrel) {
```
brutusthebee: At column 29: [(no-redeclare) 'var_idx' is already defined.](http://eslint.org/docs/rules/no-redeclare)

batavia/VirtualMachine.js
```diff
@@ -1428,6 +1495,18 @@ VirtualMachine.prototype.byte_BUILD_MAP = function(size) {
...
+    for (var i = 0; i < values.length; i += 1) {
+        dict.__setitem__(keys[i], values[i])
+    }
+    this.push(dict)
+    return
+}
+
 VirtualMachine.prototype.byte_STORE_MAP = function() {
     switch (constants.BATAVIA_MAGIC) {
         case constants.BATAVIA_MAGIC_35:
```
brutusthebee: At column 5: [(no-useless-return) Unnecessary return statement.](http://eslint.org/docs/rules/no-useless-return)

batavia/VirtualMachine.js
```diff
@@ -1818,44 +1920,82 @@ VirtualMachine.prototype.byte_CALL_FUNCTION_VAR = function(arg) {
...
-    }
+        var retval = func(posargs, namedargs)
+        this.push(retval)
+    } else {
+        var namedargs = new types.JSDict()
+        var lenKw = Math.floor(arg / 256)
+        var lenPos = arg % 256
+        for (var i = 0; i < lenKw; i++) {
+            var items = this.popn(2)
+            namedargs[items[0]] = items[1]
```
brutusthebee: At column 13: [(no-redeclare) 'namedargs' is already defined.](http://eslint.org/docs/rules/no-redeclare)

batavia/VirtualMachine.js
```diff
@@ -1818,44 +1920,82 @@ VirtualMachine.prototype.byte_CALL_FUNCTION_VAR = function(arg) {
...
+        }
+        if (kwargs) {
```
brutusthebee: At column 13: [(no-redeclare) 'lenPos' is already defined.](http://eslint.org/docs/rules/no-redeclare)

batavia/VirtualMachine.js
```diff
@@ -1818,44 +1920,82 @@ VirtualMachine.prototype.byte_CALL_FUNCTION_VAR = function(arg) {
...
+            for (let kv of kwargs.items()) {
+                namedargs[kv[0]] = kv[1]
+            }
+        }
+        var posargs = this.popn(lenPos)
+        if (args) {
+            for (let elem of args) {
+                posargs.push(elem)
+            }
+        }
```
brutusthebee: At column 13: [(no-redeclare) 'posargs' is already defined.](http://eslint.org/docs/rules/no-redeclare)

batavia/VirtualMachine.js
```diff
@@ -1818,44 +1920,82 @@ VirtualMachine.prototype.byte_CALL_FUNCTION_VAR = function(arg) {
...
+        var func = this.pop()
+        if (func.__call__ !== undefined) {
+            func = func.__call__.bind(func)
+        }
 
-    var retval = func(posargs, namedargs)
```
brutusthebee: At column 13: [(no-redeclare) 'func' is already defined.](http://eslint.org/docs/rules/no-redeclare)

batavia/VirtualMachine.js
```diff
@@ -1818,44 +1920,82 @@ VirtualMachine.prototype.byte_CALL_FUNCTION_VAR = function(arg) {
...
-    this.push(retval)
+        var retval = func(posargs, namedargs)
+        this.push(retval)
+    }
 }
 
 VirtualMachine.prototype.byte_RETURN_VALUE = function() {
```
brutusthebee: At column 13: [(no-redeclare) 'retval' is already defined.](http://eslint.org/docs/rules/no-redeclare)

abonie:
Thanks lucantrop. There were some unexpected successes, they just did not appear at the bottom.

And it turns out there already was a solution for this problem in `not_implemented_versions` :)

phildini:
This is pretty incredible, and you've got a 🚢 from me.

It's a shame we have to repeat the <3.6 cases so many times. If you felt like condensing that logic at all, that would be fascinating and most likely very helpful.

abonie:
@phildini You mean that CI task for Python 3.4 has to be run each time, even though PR regards Python 3.6?

Well, I am not well-versed with Ops part of DevOps, so one thing that comes to my mind would be parsing commit/PR message by BeeKeeper, looking for some special line like "# test Python 3.6" and changing tasks based on that. This sounds like a mess though, and also note that:
- Ultimately we would like to have separate branches for different Python versions, and then this problem is easily solved via beekeeper.yml config file.
- As long as Batavia supports different versions on a single branch, there is a risk, that accidentally someone will mess up how it behaves for Python 3.4 while working on Python 3.5, so running all tests may actually be necessary :(

phildini:
Sorry, I should have been clearer! 

What I meant was we do

```
case constants.BATAVIA_MAGIC_34:
case constants.BATAVIA_MAGIC_35a0:
case constants.BATAVIA_MAGIC_35:
case constants.BATAVIA_MAGIC_353:
...
case constants.BATAVIA_MAGIC_36:
...
```
a whole lot in these changes, which seems like duplicating a lot of code? Also maybe brittle for supporting future versions.

If there were some helper functions to deal with this, it would make me happy, but this whole complaint is more "nit" than "blocker", so feel free to say you don't want to action that right now and we can probably still move forward.

abonie:
@phildini Oh, sure. I have actually already started working on a module that would allow something more like this:
```javascript
if (version.older('3.6')) {
    ...
} else {
    ...
}
```
I thought I'd submit it in separate patch though, I am not sure if I can finish it before Monday, since Thursday through Sunday I won't be able to do any work (and it's not super straight forward as there are some caveats with pre/postreleases).

phildini:
Fair enough! If you get to the point where tests are passing on this branch, I'm happy to merge in and save the optimization for post-merge.

abonie:
Great! BTW I don't know what's going on with CI right now, seems like it crashed on Python lint check..?

phildini:
I'm not entirely sure what's going on either... could you try rebasing against master and pushing again?

phildini:
@abonie do you feel this is ready to merge? I'm happy to do so if you're ready.

batavia/builtins/bytes.js
```diff
@@ -26,7 +27,19 @@ function bytes(args, kwargs) {
     } else if (args.length === 1) {
         var arg = args[0]
         if (arg === null) {
-            throw new exceptions.TypeError.$pyclass("'NoneType' object is not iterable")
+            switch (constants.BATAVIA_MAGIC) {
```
phildini: It would be great if we could condense these cases down to a helper function, but I'm fine with that being a cleanup PR later.

abonie:
@phildini I made one last change or rather retracted one of the earlier commits, as it seems that error message in `sorted` has been changed in Python 3.6.2 (which I had on my machine), but BeeKeeper runs Python 3.6, so that change actually caused a failure. It is definitely ready to be merged. Failing test case in `test_str.py` in `FormatTests` seems like more than just an error message discrepancy.

PR #667, opened 2017-10-05
Reorganize docs to use grid layout.

freakboy3742:


freakboy3742:
Yes - you can build locally. `make html` in the docs directory. There's an outstanding task to add a beekeeper/beefore tasks for docs compilation.

glasnt:
Not sure why the builds are taking so long...

PR #671, opened 2017-10-09
Node script to run Python code using Batavia

xitij2000:
Added Node script to run Python file in a Batavia VM. The node script includes a custom loader uses a Python script to compile the module or file path. The Python script outputs the base64-encoded bytecode and file name, so it can be used the Batavia VM. This allows imports to mostly work. 

There was no simple way to type some code in a file and have it run in using the Batavia VM. This change hopes to address that. 

## PR Checklist:


compile_module.py
```diff
@@ -0,0 +1,65 @@
...
+import base64
+import importlib
+import os
+import tempfile
+
+sys.path.insert(0, os.getcwd())
+
+
+def get_module_path(module):
+    if module.endswith('.py'):
```
phildini: It seems weird to me that we have to special-case this but otherwise can use `importlib`? Why can't `importlib` be used in this case?

docs/intro/tutorial-2.rst
```diff
@@ -0,0 +1,17 @@
...
+
+Batavia includes a simple command-line script that you can use to run your
+Python code. To use this tool you need to have followed the instructions from
+:doc:`tutorial-0`.
+
+You can now run Python code from a code from the command line as follows:
+
+.. code-block:: bash
+
+    node run_in_batavia.js /path/to/python_file.py
```
phildini: The standard way to [call scripts in npm packages](https://docs.npmjs.com/cli/run-script) is `npm run {script_name}`. 

Could we add a script entry to the `package.json` so the command is something like `npm run python /path/to/python/file`?

phildini:
I think this looks pretty good, thanks for submitting.

Since we're more-or-less standardizing on npm, I'd prefer us to rely on that for running scripts, etc.

compile_module.py
xitij2000: I added this to allow someone to run a Python file as: ``./run_in_batavia.py /path/to/file.py`` where is where the execution will start. After that each time ``compile_module.py`` will be called it will be with an import path like ``package.module``. If importlib is given a full path to a Python file I don't think it will work. 

Although now that I think about it it should work if given ``file.py`` from the current directory. I will make this change and update the pull request.

docs/intro/tutorial-2.rst
xitij2000: Certainly, will add that.

phildini:
This is awesome, thanks for the changes. Once tests pass, happy to merge in.

xitij2000:
There is one issue I see, which is that running the npm script as ``npm run python`` changes the current working directory to the project directory, which means relative paths to a Python file wont work. This is why I specified absolute path in the docs. 

I don't know how to get the "real" current directory when running a script as "npm run {script}".

phildini:
Hmm... do the node path docs help at all? It looks like there's some bits about absolute paths and normalization.

xitij2000:
If you run the script directly via Node everything is fine, you can get the current directory via ``process.cwd`` or use ``path.resolve``. However if you run via npm, it changes the working directory, so you can't get the directory in which you actually run the command.

Since npm scripts can anyway only be run under the directory of the project i.e. the directory containing ``package.json`` or any directory below it this will only become a problem if 1) you are in a subdirectory inside the Batavia code checkout and 2) you pass ``npm run python`` a relative path. 

So if a create a directory called ``playground`` inside the Batavia project directory, create a Python file called ``scratch.py`` and then inside the ``playground`` directory  I run ``npm run python scratch.py`` the command will fail because it will look for ``scratch`` relative to the project root, not the current directory. Running ``node ../run_in_batavia.js scratch.py`` should work however since directly running in Node gives it access to the real cwd.

phildini:
I think it's still true that we want to be encouraging `npm run python` as the first-best way to get things done, even if that comes with some limitations for now. I think the limitations are surmountable, but since we still have `run_in_batavia`, I think this is fine to merge as-is. @xitij2000 
 Do you want to try to play with the path stuff a bit more or have me merge?

phildini:
This also might be worth looking at for how to parse file paths more easily: https://github.com/extrabacon/python-shell

xitij2000:
I explored the path stuff, and it seems like a solution to this problem landed in npm v5.4.0, an environment variable called ``INIT_CWD`` now contains the initial CWD where the npm command was run. 

I am adding code to check if this environment variable is defined, and if it is use it, or fall back on the the current behaviour.

Issue #478, opened 2017-03-30
GSOC: str, int, dict, sets, range and eval


memo330179:
# Introduction

Completion of a project is important. In order for the product to be useful, it must provide a complete set of tools to work with. As it stands, Batavia has an extensive library of data structures and data types that make up the Python programming language. However, it also has many functions missing from those data types, as well as, some functions that are not behaving in the way they are expected to. In order for the project to come to fruition, these functions must be implemented. My work would be to implement some of the string functions that are still not working in the batavia project.

# Project Goals

## .format()

the tools I use the most when I am programming in Python are those found in the string object, and, of those, the string method I use the most is string formatting. In Python the recommended way to format string is using the built-in format function.

&quot;Hi, my name is {0} and I want to contribute to  {1}&quot;.format(&quot;Guillermo&quot;, &quot;Batavia&quot;)

Following the work of the implementation of string.\_\_mod\_\_ done by Jacob Stoebel, I want to add the new formatting style. 

## Other things

One thing that makes Python so cool is the fact that most of the time if you need to do something it has already been implemented. This is the case with sorting, treating lists as stacks, etc. It makes sense then that these functions that make Python so useful get added onto batavia. The team has done an excellent job so far and I would like to push it a bit further.

There are string functions that aren&#39;t implemented most of them are relatively simple. For example, isnumeric which would be a check to see if it is a number. (see more in timeline)

Most of the list functions have been implemented, but there are a few of them that have not such as pop or insert. These functions are important to be able to treat lists as stacks or queues.

(see more in timeline)

The dictionary module is also lacking some functions such as setdefault (see more in the timeline)

Set is also missing just a couple methods that make it cooler as well as alternate ways to call existing methods

It used to be that range was a list and so you could call list methods on it, but this changed and not it is a range class. However, the usefulness of it should be the same and collection methods were added to it so that you could do things such as use the in operator or access elements using an index. I would like to implement those as well.

# Implementation

## \.format() ##

The implementation of this would follow the implementation described in PEP 3101. Some of the requirements for this project are:

- Being able to distinguish markup from data:
  - In the code &quot;My name is {0}&quot;.format(&#39;Fred&#39;), &quot;My name is&quot; would be the data and would be passed directly, whereas {0} signifies text to be changed.
  - Braces should be able to be escaped by doubling them &quot;{{}}&quot; -&gt; {}
- A developer can use a name or a string to define what the markup needs to be replaced with.
[...]

PR #684, opened 2017-11-12
Add set.union and set.intersection aliases to set.__ior__ and set.__i…

lschumm:
`set.union` and `set.intersection` are two frequently used methods of a set. Issue #478 added the `set.__ior__` and `set.__iand__` methods which are equivalent–although `.union` and `.intersection` are more standard (as they aren't dunder methods) and are used in a few modules in the python standard library.

## PR Checklist:

### DIFF
--- a/tests/datatypes/test_set.py
+++ b/tests/datatypes/test_set.py
@@ -62,6 +62,34 @@ class SetTests(TranspileTestCase):
             """)
 
 
+    def test__ior__(self):
+        self.assertCodeExecution("""
+        a = {'a', 'b', 'c'}
+        b = {'a', 'c', 'd'}
+        a.__ior__(b)
+        """)
+
+    def test__iand__(self):
+        self.assertEqual("""
+        a = {'a', 'b', 'c'}
+        b = {'a', 'c', 'd'}
+        a.__iand__(b)
+        """)
+
+
+    def test_union(self):
+        self.assertCodeExecution("""
+        a = {'a', 'b', 'c'}
+        b = {'a', 'c', 'd'}
+        a.union(b)        
+        """)
+
+    def test_intersection(self):
+        self.assertCodeExecution("""
+        a = {'a', 'b', 'c'}
+        b = {'a', 'c', 'd'}
+        a.intersection(b)
+        """)
 
 class UnarySetOperationTests(UnaryOperationTestCase, TranspileTestCase):
     data_type = 'set'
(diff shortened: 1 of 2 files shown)


tests/datatypes/test_set.py
```diff
@@ -62,6 +62,34 @@ def test_iter(self):
...
+        b = {'a', 'c', 'd'}
+        a.__iand__(b)
+        """)
+
+
+    def test_union(self):
+        self.assertCodeExecution("""
+        a = {'a', 'b', 'c'}
+        b = {'a', 'c', 'd'}
+        a.union(b)
```
freakboy3742: Although this is executes the right code, it won't actually test anything. `assertCodeExecution()` performs a comparison of the output of executed code - so, this test will verify that the code doesn't output anything. This will verify that the code runs without error, but won't verify that the output is actually correct.

To test that the output is correct, you have to print the result of the operation so that there is something to compare.

tests/datatypes/test_set.py
```diff
@@ -62,6 +62,34 @@ def test_iter(self):
...
+        a = {'a', 'b', 'c'}
+        b = {'a', 'c', 'd'}
+        a.union(b)        
+        """)
+
+    def test_intersection(self):
+        self.assertCodeExecution("""
+        a = {'a', 'b', 'c'}
+        b = {'a', 'c', 'd'}
+        a.intersection(b)
```
freakboy3742: As with `union`, a print statement (or something else that generates output) is required.

tests/datatypes/test_set.py
```diff
@@ -62,6 +62,34 @@ def test_iter(self):
             """)
 
 
+    def test__ior__(self):
+        self.assertCodeExecution("""
+        a = {'a', 'b', 'c'}
+        b = {'a', 'c', 'd'}
+        a.__ior__(b)
```
freakboy3742: The check of `__ior__` isn't needed - it's covered automatically by the InlineBinaryOperationTests.

tests/datatypes/test_set.py
```diff
@@ -62,6 +62,34 @@ def test_iter(self):
...
+        """)
+
+    def test__iand__(self):
+        self.assertCodeExecution("""
+        a = {'a', 'b', 'c'}
+        b = {'a', 'c', 'd'}
+        a.__iand__(b)
```
freakboy3742: The check of `__iand__` isn't needed - it's covered automatically by the InlineBinaryOperationTests.

freakboy3742 (changes requested):
The implementation looks right - but the test isn't actually testing what you think it is. Beef up the test, and this should be ready to go!

phildini:

Issue #602, opened 2017-08-07
PyStone seems to be going into an infinite loop


timb07:
Using the Batavia testbed, clicking "Run PyStone (5 loops)" seems to go into an infinite loop: the browser process uses 100% CPU, and the browser tab becomes unresponsive.

timb07:
Copying and pasting the PyStone code from `test_pystone.py` into the testbed and running it directly (with `main(1)`) works fine.

davem7:
I was just going through the tutorial and thought I was having the same issue, but after letting it sit for a bit, the Console output showed the results for 50000 loops.  It seems to execute 50000 loops regardless of which button is pressed.

martica:
It looks like the issue you found (List.__len__ returning a JS int) @timb07 is resolved now. Thanks for the report and the investigation.

PR #695, opened 2017-12-04
[RE]Fix pystone

### LINKED ISSUE #687: Two odd functionalities on running PyStone
Hi, I was going through its features and just playing around with it before i start contributing this project. While doing so, I found something odd while i was running PyStone on my local environment. 

Firstly, it appears to give a result of 50000 loops regardless of which loops we choose to run.
Secondly, it takes too long time (more than 100 sec) to run PyStone when it takes less than a second on my terminal.

By the way, I'm using Python 3.5 and Mac OS with High Sierra 10.13.1

Boeing737ng:
**------------------------------- Describe your changes in detail -----------------------------------**

Previously, I wrote an issue #687 (currently closed) which was duplicate of one of a comment in #602. It's about odd functionality of PyStone.

When we press the pystone button, it appears to bring _pystone.py_ file from our own PC path.
The parameters from the button ('5', '500', '50000') wasn't passing correctly to that file so the pystone only gave the benchmark result of 50000 loops which is a default value in _pystone.py_.

To fix it, I created a copy file of _pystone.py_ and located in /testserver directory(same as _sample.py_ directory) and made some change to get the parameter value which is defined as "sys.argv".
I also edited the parameter `bytecode(pystone.__file__)` to `bytecode(pystone.py)` in _urls.py_ so now  it calls the file directly from it's current directory. 

**----------------------------- What problem does this change solve? -------------------------------**

As I mentioned in a closed issue #687, Pystone buttons only appeared to give a result of "50000" loops no matter which pystone button we click. 
Now it gives a correct result depends on its given number of loops(passes by parameter).

**------------------------------------------- Fixed issue ---------------------------------------------**
It Fixes #687 

Please suggest if I need to do something else for this changes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1008,6 +1008,31 @@
       "integrity": "sha1-TJQj6i0lLCcMQbK97+/5u2tiwGo=",
       "dev": true
     },
+    "cache-loader": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/cache-loader/-/cache-loader-1.2.0.tgz",
+      "integrity": "sha512-E95knP7jxy2bF/HKuw5gCEXm06tp7/sEjewNF39ezyVBnVmNzB9bnXflEFBvrqZrswsCmgiCbiIc7xIeVXW7Gw==",
+      "dev": true,
+      "requires": {
+        "async": "2.5.0",
+        "loader-utils": "1.1.0",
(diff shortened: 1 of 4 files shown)


Boeing737ng:
I'm sorry the smoke build had been stopped due to my mistake. I will submit a new PR again.

PR #702, opened 2017-12-17
Fixed implementation Float.__floordiv__ for Int

Axle7XStriker:
Fixed the implementation for Float.__floordiv__  for division of it by integer datatype by changing the "other.valueOf()" by "other.__float__().valueOf()" as initially float datatype was directly getting divided by integer datatype according to javascript standards but now after making changes the float datatype will get divided by integer datatype according to python standards.

 Now the test_floor_divide_int for float datatype is passing which was initially marked as expected failure.

Refs #46 

## PR Checklist:

### DIFF
--- a/batavia/types/Float.js
+++ b/batavia/types/Float.js
@@ -316,7 +316,7 @@ Float.prototype.__floordiv__ = function(other) {
 
     if (types.isinstance(other, types.Int)) {
         if (!other.val.isZero()) {
-            return new Float(Math.floor(this.valueOf() / other.valueOf()))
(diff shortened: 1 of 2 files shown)


Axle7XStriker:
I have checked and I am not getting any errors in the given test which has failed in the smoke-test. 
I am using OS:- Linux Mint; Python:- 3.4.4

freakboy3742:
@Axle7XStriker From the look of it, I'm guessing you're not running the same version of Python 3.4; the CI server runs 3.4.7. CPython will sometimes change the error message that is raised between micro versions.

PR #705, opened 2018-01-05
Replace Webpack with Rollup, and start using ES2016 syntax

freakboy3742:
This is a major refactor, replacing Webpack with Rollup. 

Rollup's configuration is *much* easier to work with, and produces smaller output, faster.

The syntax maps are also much more reliable, and in the territory of "they just work", rather than requiring a whole separate webserver serving JS content.

Lastly, this patch packages Batavia as a Python package (as well as a JS one), so that end users can just "pip install batavia" and get a Python-compatible deployment. It currently only includes the pieces for a Django deployment, but it's really just a staticfiles directory, so any other deployment should also be possible.


phildini:

PR #708, opened 2018-01-10
Add some missing string methods; fix some isinstance and type checking

swenson:
* Add string methods `isidentifier`, `isalpha`, `isdigit`, `join`, `index`
* Types are supposed to have a `__dict__` `mappingproxy` -- I put in a `__dict__` entry with a dictionary for now. We don't have a `mappingproxy` yet, but we can stub it out.
* `str.split` previously returned a native array, which doesn't always play well with, say, `map`.
* Support `type()` equality checking.
* Support some more special cases of `isinstance`.
* Fix a bug in `sys.stdout` buffering that relied on weird JS behavior to output the correct result
* `list.__len__` should return a proper `Int`
* Add flakey test marker to some `dict` tests.

## PR Checklist:


swenson:
Thanks!

PR #710, opened 2018-01-15
__len__ return type fixed; isinstance() support is better

swenson:
* `__len__` should always return a Python `Int`
* `isinstance` should work for more types
* `Exception` passed to `PyErr_SetString` is a Python callable, not a JS class

## PR Checklist:


swenson:
Thanks!

PR #716, opened 2018-02-04
Account for -0 when doing int ops

swenson:
The bignumber library allows `-0`, which causes lots of weird things to happen that should not happen.

## PR Checklist:


swenson:
Thanks!

PR #718, opened 2018-02-25
Add more information to the setup.py for batavia.

phildini:
Per https://packaging.python.org/tutorials/distributing-packages/#project-urls
we can have even more helpful information in our package and on our PyPI page by
adding urls to the setup.py

## PR Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,5 +32,11 @@ setup(
         'Topic :: Software Development',
         'Topic :: Utilities',
     ],
-    test_suite='tests'
+    test_suite='tests',
+    package_urls={
+        'Funding': 'https://pybee.org/contributing/membership/',
+        'Documentation': 'http://batavia.readthedocs.io/en/latest/',
+        'Tracker': data['bugs']['url'],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -32,5 +32,11 @@
         'Topic :: Software Development',
         'Topic :: Utilities',
     ],
-    test_suite='tests'
+    test_suite='tests',
+    package_urls={
+        'Funding': 'https://pybee.org/contributing/membership/',
+        'Documentation': 'http://batavia.readthedocs.io/en/latest/',
+        'Tracker': data['bugs']['url'],
```
glasnt: A question: where is this being sourced from?

phildini: The `data` comes from the `package.json` in the repo. Batavia is semi-unique in that is has two package-management-config-files. As best we can, de-deduping between them is great. Sadly, `npm` does not yet support a donate link.

freakboy3742: The use of package.json will change as soon as I can land PR #705 - that will involve allow packaging Batavia as Python code, rather than Javascript.

PR #730, opened 2018-03-13
union and intersection frozen set

mdmjg:
These are the union and intersection methods for frozen set.

### DIFF
--- a/tests/datatypes/test_frozenset.py
+++ b/tests/datatypes/test_frozenset.py
@@ -18,19 +18,15 @@ class FrozensetTests(TranspileTestCase):
     def test_union(self):
         self.assertCodeExecution("""
         x = frozenset([2,4,5,6])
-        y = frozetset([3,4,7,9])
-        z = 5
+        y = frozenset([3,4,7,9])
         print(x.union(y))
-        print(x.union(z))
         """)
 
     def test_intersection(self):
         self.assertCodeExecution("""
-        x = frozenset([2,4,5,6])
-        y = frozetset([3,4,7,9])
-        z = 'dog'
+        x = frozenset([2,4,5,6,7])
+        y = frozenset([3,4,7,9])
         print(x.intersection(y))
-        print(x.intersection(z))
         """)
 
 
--- a/tests/datatypes/test_frozenset.py
+++ b/tests/datatypes/test_frozenset.py
@@ -15,6 +15,24 @@ class FrozensetTests(TranspileTestCase):
         print(len(frozenset([1, 2])))
         """)
 
+    def test_union(self):
+        self.assertCodeExecution("""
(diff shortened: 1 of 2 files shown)


freakboy3742 (changes requested):
This implementation looks great; now we need some tests to go with it (otherwise we have no guarantee that the code will keep working!) 

Batavia tests take the form of samples of Python code that will be executed in both CPython and Batavia; the test case then compares the output of both runs to check that they're the same. 

If you look in the datatype tests, test methods with calls to `assertCodeExecution` are what you're looking for. Add some tests like those to test frozenSet behavior, and this PR should be good to go!

tests/datatypes/test_frozenset.py
```diff
@@ -15,6 +15,24 @@ def test_len(self):
         print(len(frozenset([1, 2])))
         """)
 
+    def test_union(self):
+        self.assertCodeExecution("""
+        x = frozenset([2,4,5,6])
+        y = frozetset([3,4,7,9])
+        z = 5
+        print(x.union(y))
+        print(x.union(z))
```
freakboy3742: This is a good test, and it probably passes, but it's not the right way to test it in Batavia.

Batavia tests are based on comparing output of two program runs; and, to do that, it's necessary for both programs to terminate successfully, not raise an error.

If you're testing behavior that you know will (and should) raise an error, your test code should catch the exception and print a "raised an error" message; that way, the comparison has something to match up.

freakboy3742 (changes requested):
Looking good; one small tweak needed on the test cases, and we should be good to merge!

mdmjg:
I made some changes but there is an important issue that could be addressed in order to make these tests pass. When you do the following in CPython:
x = frozenset([5,1,7,6])
print(x)
 It will return frozenset({1, 5, 6, 7}) <- an ordered version of the frozen set.

However, when you do the same in Batavia, the program will not return the set in increasing order. This makes the union function return a frozen set that is not increasing order, which in turn makes the tests fail. In order to make the union function return exactly the same as CPython, changes unrelated to the union method would have to be implemented.

martica:
CPython isn't quite as predictable in its ordering. 

These tests could be made to pass by changing print(frozenset... to print(sorted(frozenset... if the sorted() builtin was implemented.

Barring that, it looks like changing the format to something that tests properties of the output set would cover this.

Like:
```
x = frozenset([2,4,5,6])
y = frozenset([3,4,7,9])
z = x.union(y)
print(len(z))
print([elem in z for elem in x])
print([elem in z for elem in y])

x = frozenset([2,4,5,6])
y = frozenset([3,4,7,9])
z = x.intersection(y)
print(len(z))
print([elem in x and elem in y for elem in z])
```

phildini:

Issue #735, opened 2018-03-29
Wrong error type returned by pow function


djmgit:
## Expected Behavior
Pow function should return ValueError when there are 3 arguments and the second argument is
negative.

## Current Behavior
Batavia is returning TypeError for the above mentioned case. This is also causing the tests in tests/builts/test_pow to fail.

## Steps to reproduce

1. Execite the following code snippet using Batavia
    ```
    print(pow(1, -1, 1)
    ```
    Batavia returns TypeError:
    ```
    Traceback (most recent call last):
    File "/home/deep/batavia/test3.py", line 1, in <module>
    TypeError: pow() 2nd argument cannot be negative when 3rd argument specified
    ```
    however python returns ValueError
    ```
    Traceback (most recent call last):
    File "../test3.py", line 1, in <module>
    print(pow(2, -4, 1))
    ValueError: pow() 2nd argument cannot be negative when 3rd argument specified
    ```
## Your Environment

* Python Version (list the specific version number)
3.5.2
* Operating System and Version (select from the following and list the specific version number; if your OS is not listed, list that as well)

  - [ ] OSX
  - [ ] Linux
  - [ ] Windows
  - [ ] Ubuntu

djmgit:
I am working on this.

martica:
This was fixed by #784.

PR #736, opened 2018-03-29
Fix issue #735 - Wrong error type returned by pow function

djmgit:
Fixed issue #735. Pow function now returns ValueError
instead of TypeError when it has got 3 arguments and
second one is negative.

## PR Checklist:


martica:
Thanks for the fix, @djmgit. It looks like that error type was different in different python versions. We've submitted a fix for this now. (I didn't realize you had an open PR for it when I started on the project at PyCon. Didn't mean to step on your toes.)

PR #740, opened 2018-04-21
issubclass() builtin implemented

marcus290:
Implemented issubclass() builtin method, using the existing types.issubclass method.

Implemented the missing issubclass() builtin method

Complete implementations of Python builtins #47

## PR Checklist:

- [x] All new features have been documented (documentation for builtins are already in the existing code)


marcus290:
Hi there. I added some tests - sorry about that!

But I found an issue because Python has a quirk where bool is recognized as a subclass of int. But currently Batavia does not.

I suspect we would need to change something in Batavia's implementation of bool type, instead of me adding a hack to this builtin or the existing types.issubclass method. 

Should I remove these tests and open a new issue?

marcus290:
Tests added! 

All fixed.

Revised types.issubclass() in types.js, which previously (incorrectly) processed an instance object and a class as parameters, instead of two classes.

martica:
@freakboy3742 These changes look good to go in. Are you able to approve the changes that added tests?

PR #743, opened 2018-05-14
Implement integer pow operation with complex operands

EtiennePelletier:
### I am implementing the pow operation for standard type integer with complex operands.

Also enabled int pow float tests as the code is already implemented and working, as well as added a clarification at the end of the execution of sample.py to remove the ambiguity with the last line displaying 'Error' in the console, even if that is expected.

The added tests were failing initially because Batavia had imprecisions with very small numbers (for example showing 9.1535235e-38 instead of 0), as well as inconsistently displaying +0j and -0j. I was able to fix both of these, required to make the int ** complex tests pass.

I also uniformized Float (and Complex) object's display (__str__) to match Python (when to use exponential form, how it is used, when to or not to round, etc.)

All the tests pass! I can probably remove some more tests from the not_implemented lists, but I will take a look at that after the current PR gets merged (and updated if required).

#46 

## PR Checklist:


EtiennePelletier:
@danyeaw It looks like the builds are hung... I'm waiting on knowing if all tests pass before checking my last checkbox and removing the "WIP" from PR title.
```
Status
Error: Unable to start worker: AGENT
```

https://beekeeper.herokuapp.com/projects/pybee/batavia

EtiennePelletier:
The tests aren't passing, because JavaScript's scientific notation for floats isn't implemented like Python's yet.
Once done, I will review what other tests I can take off the not_implement lists, there will be a few others.

batavia/types/Bool.js
```diff
@@ -299,7 +307,7 @@ Bool.prototype.__pow__ = function(other) {
             }
```
brutusthebee: At column 46: [(eqeqeq) Expected '===' and instead saw '=='.](http://eslint.org/docs/rules/eqeqeq)

EtiennePelletier:
The tests are now all passing! @danyeaw and @freakboy3742 , this PR is ready for review 😃 

Note: It was my first time using Node.js, so there is one function I created in Float.js that I needed to reuse in Complex.js, so maybe my way of importing it is not the best one. Please let me know if you think about a better design.

PR #746, opened 2018-05-15
fix: Ensure `pow(3, -4, 5)` throws ValueError

cLupus:
`pow(x, y, z)` throw a `ValueError` when `y` is negative, while `x`, and ,`z` are positive.

Additionally cleaned up the javascript with ECMAscript 6 recommendations; i.e. the use of `let`, or `const` instead of `var`, and triple `=` instead of double.

Related issue: #46

## PR Checklist:

### DIFF
--- a/batavia/builtins/pow.js
+++ b/batavia/builtins/pow.js
@@ -1,8 +1,8 @@
-var exceptions = require('../core').exceptions
-var types = require('../types')
+const exceptions = require('../core').exceptions
+const types = require('../types')
 
 function pow(args, kwargs) {
-    var x, y, z
+    let x, y, z
     if (!args) {
         throw new exceptions.TypeError.$pyclass('pow expected at least 2 arguments, got 0')
     }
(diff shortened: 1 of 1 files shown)


phildini:
Hi there! Did this fix any tests? If not, should tests be added to prove the fix works?

cLupus:
Do you have a suggestion on how to write a test for that?

Note that, the previous two tests (`pow(3, -4, 5)`, and some other `pow` with a negative power, and a modulo) failed before this PR.

shaib:
Well, I can't seem to be able to approve the PR, but I can verify that when I run the tests on Python 3.6, there are two failing tests in `tests.builtins.test_pow.PowTests` -- `test_int_neg_y_pos_z` and `test_int_neg_y_neg_z` -- which fail without the PR, and pass with it; and as far as I can see, this is exactly what the PR set out to fix.

freakboy3742:
@shaib Hrm - that's odd - any idea why those tests aren't failing on CI?

martica:
Thanks for the PR, @cLupus! It looks like this got covered by another PR, so I'm going to close this out.

PR #752, opened 2018-05-26
Add recommentation to use nodeenv to install node

shaib:
This is my preferred way of having node.js in a Python project, so I'm suggesting it.

- [X] All new features have been tested
- [X] All new features have been documented
- [X] I have read the **CONTRIBUTING.md** file
- [X] I will abide by the code of conduct

### DIFF
--- a/docs/how-to/contribute-code.rst
+++ b/docs/how-to/contribute-code.rst
@@ -52,8 +52,17 @@ Install Node.JS
 ^^^^^^^^^^^^^^^
 
 Lastly, you'll need to install `Node.js`_. You need to have a recent version
-of Node; we test using v6.9.1. Once you've installed node, you can use it to
-install Batavia's JavaScript dependencies, and compile the Batavia library:
+of Node; we test using v6.9.1. One civilized way to get an isolated Node.js
+environment is using the `nodeenv`_ package -- this ties your Node.js
+environment into your Python virtual environment.
+
+.. code-block:: bash
+
+    $ pip install nodeenv
+    $ nodeenv -p
+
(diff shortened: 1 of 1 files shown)


docs/how-to/contribute-code.rst
```diff
@@ -52,8 +52,17 @@ Install Node.JS
 ^^^^^^^^^^^^^^^
 
 Lastly, you'll need to install `Node.js`_. You need to have a recent version
-of Node; we test using v6.9.1. Once you've installed node, you can use it to
-install Batavia's JavaScript dependencies, and compile the Batavia library:
+of Node; we test using v6.9.1. One civilized way to get an isolated Node.js
```
freakboy3742: "civilized" is a little pejorative; also, if we're going to make the recommendation, the docs should commit to it - "We suggest using `nodeenv`", rather than "one way is to...".

Should we also add the `--node=6.9.1` flag to ensure the right version is available?

freakboy3742 (changes requested):

docs/how-to/contribute-code.rst
shaib: I have a hard time committing to this suggestion, because my experience with node.js is rather limited. I agree to a less opinionated phrasing, we can drop "civilized".

With respect to the node version -- adding a specific one sounds like a good idea, but I've been working with the default (latest, currently v10.2.1) and there don't seem to be issues. Should we perhaps update?

rayrrr:
I just evaluated `nodeenv` and `nodenv` while getting set up to run Batavia locally and decided on the latter. Any strong reasons to choose the former instead? I'm happy so far.

phildini:

PR #754, opened 2018-05-27
Enabling test_setattr in test_list.py

shaib:
Just repeating #497 to enable it to be tested by BeeKeeper.

Once the test passes, merge that and close this.


martica:
Looks like this didn't have the desired effect. :( I'm going to close it out. #497 causes some test failures if I try it locally now, so I'll put a comment there to that effect.

PR #756, opened 2018-06-10
Implemented power and "&" operator for bytearray

MrEx88:
Signed-off-by: Jon Wesneski <jonwes2@gmail.com>


MrEx88:
This is my first contribution, so let me now if I've missed something. I see the smoke test failed; but it looks like it failed in other areas; not because of my commit.

martica:
Thanks for this! I'm going to have to close this because this patch is obsolete now. These ended up getting fixed in a different way.

PR #770, opened 2019-01-21
Update Django to Latest and Make CodeMirror npm Installed

danyeaw:
Updates Django to version 2.1.5, including:
- Fix errors caused by pyc payloads not properly decoded
- Replace MIDDLEWARE_CLASSES since now deprecated
- Upgrade urlparsers from url to the new path syntax

Since the version of CodeMirror being packaged was very old, moved it to be npm installed:
- Added it to the webpack to create a codemirror.js and codemirror.css that is placed in the static site for just the 3 files out of CodeMirror needed for this project.

Update Babel configurations and package.json to fix issues caused by the upgrade to Babel 7

## PR Checklist:

- [X] I have read the **CONTRIBUTING.md** file
- [X] I will abide by the code of conduct


danyeaw:
@freakboy3742 Thanks for your review and feedback!

After spending quite a bit of effort tracking it down, I think I resolved the issue with the builds failing hard. Webpack 4 has an issue with building for universal targets: https://github.com/webpack/webpack/issues/6525. I made use of the workaround to create a globalObject in the webpack config. 

For dependencies and building I updated it with:
- Development: `npm install` and `npm run dev` will install and build everything (no minify), including codemirror
- Production: `npm install --production` and `npm run build` will install just the normal dependencies and build a minifyed version

Running Batavia tests requires the development dependencies if PRECOMPILE is set. I thought this is much more useful for debugging while testing, but there might be different use cases. Is this what you had in mind as well?

I was not able to selectively load plugins for webpack using a single config file. If we want to reduce the number of normal dependencies then I would need to create separate dev and prod config files, and I'm not sure if we wanted that complexity.

PR #774, opened 2019-04-09
Update setup.py

zimmah:
Expands allowed versions to include python 3.6

Updated python requires and classifiers to include 3.6

It allows the use of python 3.6 and versions of 3.5 beyond 3.5.0

## PR Checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ setup(
     author_email='russell@keith-magee.com',
     url=data['homepage'],
     packages=find_packages(exclude=['docs', 'tests']),
-    python_requires='>=3.4, <=3.5',
+    python_requires='>=3.4, <=3.6',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -19,7 +19,7 @@
     author_email='russell@keith-magee.com',
     url=data['homepage'],
     packages=find_packages(exclude=['docs', 'tests']),
-    python_requires='>=3.4, <=3.5',
+    python_requires='>=3.4, <=3.6',
```
danyeaw: If we put <=3.6, I think we will get in to the situation where 3.6.1 is not allowed. What if instead we say <3.7 instead of <=3.6?

zimmah: That is a good idea

PR #779, opened 2019-05-08
Represent operators with a class.

martica:
Each operator can have a fallback, inplace falling back to normal, normal falling back to right-hand with arguments swapped.

Magic Methods now return NotImplemented instead of throwing an unsupported arguments exception, leaving the error handling to the operator class. This mimics the behaviour of the CPython implementations and simplified fall back.

Future dreams here are the hope that this will unlock implementing a more general `__dir__` that uses the methods defined directly instead of hard-coding the list.

## PR Checklist:


tests/datatypes/test_bytearray.py
```diff
@@ -242,248 +115,53 @@ class BinaryBytearrayOperationTests(BinaryOperationTestCase, TranspileTestCase):
...
+        'test_modulo_int': ['3.5', '3.6'],
+        'test_modulo_list': ['3.5', '3.6'],
+        'test_modulo_None': ['3.5', '3.6'],
+        'test_modulo_NotImplemented': ['3.5', '3.6'],
+        'test_modulo_range': ['3.5', '3.6'],
+        'test_modulo_set': ['3.5', '3.6'],
+        'test_modulo_slice': ['3.5', '3.6'],
+        'test_modulo_str': ['3.5', '3.6'],
+        'test_modulo_tuple': ['3.5', '3.6'],
+    }
```
freakboy3742: I'm guessing this is the result of 3.5 introducing string interpolation for bytearray?

tests/structures/test_datamodel.py
```diff
@@ -576,7 +572,7 @@ def __rmul__(self, other):
             print(3 * p1)
         """, run_in_function=False)
 
-    @expected_failing_versions(['3.4'])
+    @expectedFailure
```
freakboy3742: What's the underlying cause of this new (or, at least, more expanded) failure?

batavia/builtins/sum.js
```diff
@@ -1,7 +1,12 @@
+var operators = require('../operators')
 var exceptions = require('../core').exceptions
 var types = require('../types')
 var type_name = require('../core').type_name
 
+function add(left, right) {
+    return operators['__add__'].apply(left, right)
```
freakboy3742: Given this will be doing a dynamic attribute lookup on every `add`, would it be worth caching the attribute lookup as a module-level global? It wouldn't be the cleanest code, but it's one of those areas where a very small bit of dirt in the internals, could end up having a profound overall performance impact.

tests/datatypes/test_bytearray.py
martica: Yeah. These are the only behaviours on Bytearray that aren't in place yet.

tests/structures/test_datamodel.py
martica: This does look like an expansion, but it is actually a tightening. The wrapping class of this set of tests was all annotated as expectedFailure, but for some reason 3.4 wasn't respecting that (I added those 3.4 only failures to each test in this class earlier). I've removed the expectedFailure from the class now and applied it to just this one method. It does seem like an issue that 3.4 didn't respect the class being marked as expectedFailure, but hopefully Dan cuts 3.4 out of the picture shortly.

batavia/builtins/sum.js
martica: Good call.

tests/datatypes/test_range.py
```diff
@@ -1,4 +1,5 @@
-from .. utils import TranspileTestCase, UnaryOperationTestCase, BinaryOperationTestCase, InplaceOperationTestCase
+from ..utils import TranspileTestCase, UnaryOperationTestCase, BinaryOperationTestCase, InplaceOperationTestCase, \
+    MagicMethodFunctionTestCase
 
 import unittest
```
brutusthebee: At column 1: (F401) 'unittest' imported but unused

tests/datatypes/test_range.py
```diff
@@ -271,335 +240,22 @@ class BinaryRangeOperationTests(BinaryOperationTestCase, TranspileTestCase):
...
-        'test_xor_str',
-        'test_xor_tuple',
+        # Incorrect error message shown (unsupported operands vs can't multiply sequence by non-int)
+        "test_multiply_bytearray",
+        "test_multiply_bytes",
     ]
 
 
+
 class InplaceRangeOperationTests(InplaceOperationTestCase, TranspileTestCase):
```
brutusthebee: At column 1: (E303) too many blank lines (3)

tests/datatypes/test_bool.py
```diff
@@ -1,4 +1,5 @@
-from .. utils import TranspileTestCase, UnaryOperationTestCase, BinaryOperationTestCase, InplaceOperationTestCase
+from ..utils import TranspileTestCase, UnaryOperationTestCase, BinaryOperationTestCase, InplaceOperationTestCase, \
+    MagicMethodFunctionTestCase
 
 import unittest
```
brutusthebee: At column 1: (F401) 'unittest' imported but unused

tests/datatypes/test_bool.py
```diff
@@ -19,19 +20,72 @@ def test_getattr(self):
...
 
 class BinaryBoolOperationTests(BinaryOperationTestCase, TranspileTestCase):
     data_type = 'bool'
 
-    not_implemented = [
-    ]
-
 
 class InplaceBoolOperationTests(InplaceOperationTestCase, TranspileTestCase):
     data_type = 'bool'
```
brutusthebee: At column 1: (W391) blank line at end of file

tests/datatypes/test_NotImplemented.py
```diff
@@ -24,19 +29,16 @@ class BinaryNotImplementedOperationTests(BinaryOperationTestCase, TranspileTestC
...
-        'test_floor_divide_complex',
-
-        'test_modulo_complex',
+        # Incorrect error message shown (unsupported operands vs can't multiply sequence by non-int)
+        "test_multiply_bytearray",
+        "test_multiply_bytes",
     ]
 
-
 class InplaceNotImplementedOperationTests(InplaceOperationTestCase, TranspileTestCase):
```
brutusthebee: At column 1: (E302) expected 2 blank lines, found 1

tests/datatypes/test_int.py
```diff
@@ -1,4 +1,5 @@
-from .. utils import TranspileTestCase, UnaryOperationTestCase, BinaryOperationTestCase, InplaceOperationTestCase
+from ..utils import TranspileTestCase, UnaryOperationTestCase, BinaryOperationTestCase, InplaceOperationTestCase, \
+    MagicMethodFunctionTestCase
 
 import unittest
```
brutusthebee: At column 1: (F401) 'unittest' imported but unused

tests/datatypes/test_bytes.py
```diff
@@ -92,10 +92,71 @@ def test_decode_error(self):
...
+        "test_rmodulo_range": ['3.4'],
+        "test_rmodulo_set": ['3.4'],
+        "test_rmodulo_slice": ['3.4'],
+        "test_rmodulo_str": ['3.4'],
+        "test_rmodulo_tuple": ['3.4'],
+    }
+
+
+
+class UnaryBytesOperationTests(UnaryOperationTestCase, TranspileTestCase):
```
brutusthebee: At column 1: (E303) too many blank lines (3)

martica:
Adding tests for the magic methods directly to ensure that they are only defined when they should be and return NotImplemented when expected was a great idea. This has added a few more tests to the suite, but they should be useful. I have tried to get the majority of them to pass, but there are still about 4% to tidy up. Mostly in the right-hand operator implementations.

I do have one regression where the exception message has changed in some cases when multiplying a sequence by a non-int. It's the right exception type still, but the message is more generic than before. I've marked these test cases with a note about this. I'll probably have to dig into cpython to see how they accomplish the current state of those messages.

tests/utils.py
```diff
@@ -984,6 +984,117 @@ def wrapper(*args, **kwargs):
...
+
+        return func
+
+    @staticmethod
+    def _add_tests(vars, data_type_class):
+        for magic_method in MagicMethodFunctionTestCase.ALL_MAGIC_METHODS:
+            if magic_method in dir(data_type_class):
+                for datatype, examples in SAMPLE_DATA.items():
+                    test_name = 'test%s%s' % (magic_method, datatype)
+                    vars[test_name] = MagicMethodFunctionTestCase._magic_method_test(test_name, magic_method, examples[:2], small_ints=True)
```
brutusthebee: At column 120: (E501) line too long (140 > 119 characters)

tests/utils.py
```diff
@@ -984,6 +984,117 @@ def wrapper(*args, **kwargs):
...
+            else:
+                vars['test%smissing' % magic_method] = MagicMethodFunctionTestCase._magic_method_missing_test(magic_method)
```
brutusthebee: At column 120: (E501) line too long (123 > 119 characters)

tests/datatypes/test_bool.py
```diff
@@ -1,6 +1,5 @@
-from .. utils import TranspileTestCase, UnaryOperationTestCase, BinaryOperationTestCase, InplaceOperationTestCase
-
-import unittest
+from ..utils import TranspileTestCase, UnaryOperationTestCase, BinaryOperationTestCase, InplaceOperationTestCase, \
```
brutusthebee: At column 1: (F401) '..utils.SAMPLE_DATA' imported but unused

tests/datatypes/test_dict.py
```diff
@@ -279,11 +284,22 @@ class BinaryDictOperationTests(BinaryOperationTestCase, TranspileTestCase):
...
 
-    not_implemented = []
+    not_implemented = [
+        # Incorrect error message shown (unsupported operands vs can't multiply sequence by non-int)
+        "test_multiply_bytes",
+        "test_multiply_list",
+        "test_multiply_str",
+        "test_multiply_tuple",
+    ]
+
```
brutusthebee: At column 1: (W391) blank line at end of file

martica:
As we discussed, I've updated the magic method tests to only run one tests for methods that shouldn't be defined, but run a full suite for the ones that should be defined to ensure that each type is handled appropriately. This adds a more modest 2512 tests instead of the previous 9792 tests.

batavia/types/Bytearray.js
```diff
@@ -197,6 +197,16 @@ Bytearray.prototype.__rmul__ = function(other) {
     return this.__mul__(other)
 }
 
+Bytearray.prototype.__rmod__ = function(other) {
+    let types = require('../types')
```
freakboy3742: Isn't this let/require a no-op?

batavia/types/Complex.js
```diff
@@ -444,6 +448,25 @@ Complex.prototype.__rmod__ = function(other) {
     throw new exceptions.TypeError.$pyclass("can't mod complex numbers.")
 }
 
+Complex.prototype.__rmul__ = function(other) {
+    return this.__mul__(other)
+}
+
+Complex.prototype.__rpow__ = function(other) {
+    var types = require('../types')
```
freakboy3742: Isn't this the var/require a no-op?

batavia/types/Complex.js
```diff
@@ -444,6 +448,25 @@ Complex.prototype.__rmod__ = function(other) {
...
+    return new types.NotImplementedType()
+}
+
+Complex.prototype.__rsub__ = function(other) {
+    var types = require('../types')
+    return new types.NotImplementedType()
+}
+
+Complex.prototype.__rtruediv__ = function(other) {
+    var types = require('../types')
```
freakboy3742: Isn't this the var/require a no-op?

batavia/types/Complex.js
```diff
@@ -444,6 +448,25 @@ Complex.prototype.__rmod__ = function(other) {
...
+    return this.__mul__(other)
+}
+
+Complex.prototype.__rpow__ = function(other) {
+    var types = require('../types')
+    return new types.NotImplementedType()
+}
+
+Complex.prototype.__rsub__ = function(other) {
+    var types = require('../types')
```
freakboy3742: Isn't this the var/require a no-op?

batavia/types/Bytearray.js
```diff
@@ -197,6 +197,16 @@ Bytearray.prototype.__rmul__ = function(other) {
...
+    return new types.NotImplementedType()
+}
+
+Bytearray.prototype.__rmul__ = function(other) {
+    let types = require('../types')
```
freakboy3742: Isn't this let/require a no-op?

batavia/types/Float.js
```diff
@@ -504,6 +501,43 @@ Float.prototype.__getitem__ = function(other) {
...
+/**************************************************
+ * Right-hand operators
+ **************************************************/
+
+Float.prototype.__radd__ = function(other) {
+    return this.__add__(other)
+}
+
+Float.prototype.__rfloordiv__ = function(other) {
+    var types = require('../types')
```
freakboy3742: Same again here (and in other methods) - Isn't this var/require a no-op?

tests/datatypes/test_tuple.py
```diff
@@ -180,6 +181,44 @@ def test_len(self):
         """)
 
 
+class MagicMethodFunctionTests(MagicMethodFunctionTestCase, TranspileTestCase):
+    data_type = 'tuple'
+    MagicMethodFunctionTestCase._add_tests(vars(), tuple)
```
freakboy3742: This "class setup" pattern is a little incongruous with the way the other operator transpile tests are configured - is there reason this approach is needed?

batavia/types/Bytearray.js
martica: Importing types at the top level creates a circular dependency here, but in general I'm importing it just for NotImplementedType so I'll look at importing that at the top level as it comes from core, originally.

tests/datatypes/test_tuple.py
martica: It is a bit different. I'm trying to create different sets of tests depending on the properties of the python type being tested. If I create the tests at the parent class level like the other tests do, I don't have knowledge of type yet and can't select which tests are correct. I'll take another run at it and see if I can make it more straightforward to use.

batavia/types/Bytearray.js
martica: I've swapped all these out to import the NotImplemented singleton from core and imported it at the top level for all usages.

tests/datatypes/test_tuple.py
martica: I'm not seeing an easy way to straighten it out. Do you think it is an issue to do it this way?

freakboy3742: Ah - I hadn't caught on that this was in aid of doing a different construction based on the type being added. There's possibly something nifty that could be done with metaclasses, but then we'd spend as much time fixing the metaclass as we'd do using the darn thing - so this "longhand" version seems fine to me.

PR #786, opened 2019-05-17
Fix isinstance for most types.

martica:
This updates all the builtin functions for types (except memoryview) to point to the types themselves, and moves the custom creation logic to `<type>.__call__`.

`type(1) == int` is now true
`isinstance(1, int)` is now true

I've manually added test classes to test_isinstance.py for each of these types against all the sample data. I'd like to add types as a new category of sample data and have the two arg isinstance test class cover these, but it introduces a whole mess of new test failures so it seemed cleaner to break that out to its own change.

## PR Checklist:


danyeaw:
Thanks @martica! :tada:

Issue #792, opened 2019-05-23
Improve confidence in tests by enforcing that exception inside the tests are caught


martica:
It is possible to have a transpilation test pass if line in the test throws an exception. As long as those exceptions match between JS and Python, that is considered a success. The behaviour will mask issues in tests and turn a test with a error in its expression into a passing test.

```
a = 1 + "2"
print(a)
print("Done.")
```

This will pass, because both JS and Python will throw a type error about the first line and they will match.

We should catch this exception and fail the test (and fix the tests that currently violate it).

freakboy3742:
Oh yeah - this is a big one. We're very much relying on "well written tests" to catch this edge case, which is a definite footgun. 

The variant on this theme is "a test function that outputs nothing" - even if the behavior is completely different, the test will pass because there's no output, and no output == no output on both sides.

martica:
I'm most of the way through correcting tests that don't catch their exceptions. Most are okay after adding the try, but there are some that have been hiding broken behaviour. A bunch in the tests of the time module. It wasn't importing time in most of the tests and throwing matching exceptions about that. :/

PR #799, opened 2019-05-27
Add check to assertCodeExecution that something was output and fix tests

martica:
This updates the comparison logic after testing in JS and Python to include a check for both being empty.

There were a few tests that failed this check, but, happily, no test failure was hidden by this.

## PR Checklist:

- [ ] I have read the **CONTRIBUTING.md** file
- [ ] I will abide by the code of conduct

### DIFF
--- a/tests/modules/test_json/test_decoding.py
+++ b/tests/modules/test_json/test_decoding.py
@@ -1,3 +1,5 @@
+import unittest
+
 from ...utils import TranspileTestCase, ModuleFunctionTestCase, adjust
 from .JSON_data import pass_data, fail_data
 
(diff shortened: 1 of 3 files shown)


tests/modules/test_json/test_decoding.py
```diff
@@ -122,9 +125,9 @@ class LoadTests(ModuleFunctionTestCase, TranspileTestCase):
     not_implemented_versions = {
     }
 
-    not_implemented = [
-        'test_fail',
-    ]
+    not_implemented = {
+        'test_fail'
```
freakboy3742: A minor thing worth flagging as "future protection" - once [this PR](https://github.com/python/black/pull/826) is merged and released, it's quite likely I'll be convinced to turn on Black. That PR means that the trailing comma on a non-tuple is considered a significant indicator that the content is a "list of items", and shouldn't be collapsed into a single line.

martica: This was the result of a merge. I'll revert it.

martica:
fixes #792

PR #772, opened 2019-01-24
Add Initial Support for Python 3.7 and 3.8, Remove Support for 3.4

danyeaw:
Adds initial support for Python 3.7 and 3.8:
- Adds magic numbers for 3.7 and 3.8
- Add support for PEP 552 which uses 4 32-bit words in the pyc header in Python 3.7 and up
- Adds all opcodes in the virtual machine, special support for the opcodes has not been added yet
- Update the docs to have users use Python 3.5 and 3.6, and Node 10.x LTR

- The testserver runs with Python 3.7/3.8 and Node 10.x
- Some tests pass, but support for some of the new opcodes like LOAD_METHOD will come in follow on PRs.

Thanks to @freakboy3742 and @dgelessus for their help and support on gitter as I start to get my feet wet :heart:

## PR Checklist:

- [ ] All new features have been tested
- [ ] All new features have been documented
- [X] I have read the **CONTRIBUTING.md** file
- [X] I will abide by the code of conduct

PR #802, opened 2019-05-30
Replace version.earlier and version.later with version.at_least

martica:
I'm not going to merge this until Dan is finished with #772. It'd be too disruptive. But the current system is too confusing for me.

## PR Checklist:

- [ ] I have read the **CONTRIBUTING.md** file
- [ ] I will abide by the code of conduct


martica:
Dan's PR with 3.7 support is in. This is merged against that now and ready to go.

batavia/core/version.js
```diff
@@ -36,11 +36,24 @@ version.version_id = function(str) {
...
+    if (constants.BATAVIA_MAGIC === undefined) {
+        throw new exceptions.BataviaError.$pyclass('Version test called before version defined.')
+    }
     return magic_map[constants.BATAVIA_MAGIC] < version.version_id(str)
 }
 
 version.later = function(str) {
+    if (constants.BATAVIA_MAGIC === undefined) {
+        throw new exceptions.BataviaError.$pyclass('Version test called before version defined.')
+    }
```
freakboy3742: Are earlier and later still being used anywhere? Is there any reason to keep them? Is there a need to have both "later" *and* "at_least"? (I agree at_least is a much easier to read name...)

martica: I was thinking about simultaneous in-flight PRs, but it looks pretty clean right now. I'll chuck them out now and simplify things.

martica: Hah. We introduced a bunch of new ones, obviously, in the 3.7 branch. Merging against that isn't sufficient to update those as well.

PR #804, opened 2019-06-01
Fix "".format()

awildtechno:
Updated the implementation and test for new-style string formatting so that most tests pass. There is one remaining issue: Floating points are being rounded by the test suite, meaning {e} formats on floating-point numbers display incorrectly in the test suite. Someone else seems to have found a fix for that, so I should be able to take care of it as soon as I get more details.

All other pending issues appear to be fixed.

## PR Checklist:

- [ x ] All new features have been tested
- [ x ] All new features have been documented (no new features)
- [ x ] I have read the **CONTRIBUTING.md** file
- [ x ] I will abide by the code of conduct

PR #806, opened 2019-06-02
Fix test_str::test_conversion_types

awildtechno:
- Updated the test for test_str.NewFormat with @transforms and a few additional bits.
- Above change ensures that the test now accurately reflects that the .format() function should provide an exact duplicate of the Python output and Python errors.
- @tranforms was very helpful to me, but having to hunt through the file wasn't fun, so I added a detailed docstring using the existing comments.
- There was a restriction on float_exp in @transforms; however, the option was available and it was causing python output not to match up correctly. I removed the restriction.
- fixed: 'n' format code was grouped with 'g' format code, but has slightly different behavior in one case.
- fixed: both 'n' and 'g' had issues selecting the right floating point precision.
- fixed: .format() would crash with a single argument as it was expecting a Tuple.

Str.format() with new-style formatting arguments a la https://docs.python.org/3.4/library/string.html#format-string-syntax now passes muster.

Replaces PR #804

## PR Checklist:

- [ x ] All new features have been tested
- [ x ] All new features have been documented
- [ x ] I have read the **CONTRIBUTING.md** file
- [ x ] I will abide by the code of conduct

### DIFF
--- a/batavia/types/StrUtils.js
+++ b/batavia/types/StrUtils.js
@@ -1368,7 +1368,6 @@ function _new_subsitute(str, args, kwargs) {
                 break
             default:
                 throw new exceptions.ValueError.$pyclass(`Unknown format code '${type}' for object of type '${type_name(this.arg)}'`)
-
         } // switch
 
         // determine sign
--- a/batavia/types/StrUtils.js
+++ b/batavia/types/StrUtils.js
@@ -1235,12 +1235,9 @@ function _new_subsitute(str, args, kwargs) {
             throw new exceptions.ValueError.$pyclass(`Unknown format code '${type}' for object of type 'float'`)
         }
 
-        if (this.type === 'c' && ) {
+        if (this.type === 'c') {
             if (this.sign) throw new exceptions.ValueError.$pyclass("Sign not allowed with integer format specifier 'c'")
             if (this.alternate) throw new exceptions.ValueError.$pyclass("Alternate form (#) not allowed with integer format specifier 'c'")
-        }
-
-        if (this.type === 'c' && this.alternate) {
         }
 
         if (this.grouping && !type.match(/[deEfFgG%]/)) {
--- a/batavia/types/StrUtils.js
+++ b/batavia/types/StrUtils.js
@@ -287,7 +287,7 @@ function _substitute(format, args) {
 
                 if (/[diouxX]/.test(conversion)) {
                     if (!types.isinstance(arg, [types.Int, types.Float])) {
-                        throw new exceptions.TypeError.$pyclass(`%${conversion} format: a number is required, not str`)
+                        throw new exceptions.TypeError.$pyclass(conversion + "format: a number is required, not str")
                     }
                 } else if (/[eEfFgG]/.test(conversion)) {
                     if (!types.isinstance(arg, [types.Float, types.Int])) {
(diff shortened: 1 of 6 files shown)


batavia/types/StrUtils.js
```diff
@@ -1233,11 +1230,15 @@ function _new_subsitute(str, args, kwargs) {
...
         // TODO: need to check for decimal once it is implemented
         if (types.isinstance(this.arg, [types.Float]) && /[bcdoxX]/.test(type)) {
-            throw new types.ValueError.$pyclass(`Unknown format code '${type}' for object of type '${type_name(this.arg)}'`)
+            throw new exceptions.ValueError.$pyclass(`Unknown format code '${type}' for object of type 'float'`)
         }
 
-        if (this.type === 'c' && this.sign) {
-            throw new exceptions.ValueError.$pyclass("Sign not allowed with integer format specifier 'c'")
+        if (this.type === 'c') {
+            if (this.sign) throw new exceptions.ValueError.$pyclass("Sign not allowed with integer format specifier 'c'")
```
martica: Can you change these ifs to use braces for the blocks? It's not in the .eslintrc, but it is an oversight.

batavia/types/StrUtils.js
```diff
@@ -1233,11 +1230,15 @@ function _new_subsitute(str, args, kwargs) {
         // error for converting floats with improper presentation types
         // TODO: need to check for decimal once it is implemented
         if (types.isinstance(this.arg, [types.Float]) && /[bcdoxX]/.test(type)) {
-            throw new types.ValueError.$pyclass(`Unknown format code '${type}' for object of type '${type_name(this.arg)}'`)
+            throw new exceptions.ValueError.$pyclass(`Unknown format code '${type}' for object of type 'float'`)
```
martica: The .eslintrc disallows template strings. You didn't introduce this, and because the eslint submit checks aren't working this won't be flagged, but if you could convert it to a regular string with concatenation, that'd cool.

martica:
The 3.4 tests are failing because of a difference in error messages, but if you sync your PR to head, the 3.4 tests won't run anymore because we've removed 3.4 support.

awildtechno:
I think that worked, thanks again for all of your help!

martica:
Nice. All the tests are passing. Let me know if any of my comments are too cryptic.

batavia/types/StrUtils.js
```diff
@@ -584,7 +583,7 @@ function _substitute(format, args) {
 
                 case ('r'):
                     if (types.isinstance(conversionArgRaw, types.Str)) {
-                        conversionArg = `'${conversionArgValue}'`
+                        conversionArg = String(conversionArgValue)
```
martica: This actually caused a regression on the 3.6 tests. The output went from `'string'`  to `string`.

martica: %r should be format with repr() and %s below should be format with str() so both the 'r' and 's' cases here could be replaced with:

```
                case ('r'):
                    conversionArg = callables.call_method(conversionArgRaw, '__repr__')
                    break
                case ('s'):
                    conversionArg = callables.call_method(conversionArgRaw, '__str__')
                    break
```

martica: Or, even better because it will handle JS types as well as Python:

```
                case ('r'):
                    var builtins = require('../builtins')
                    conversionArg = builtins.repr([rawValue], {})
                    break
                case ('s'):
                    var builtins = require('../builtins')
                    conversionArg = builtins.str([rawValue], {})
                    break
```

package-lock.json
```diff
@@ -17,7 +17,6 @@
       "version": "7.2.2",
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.2.2.tgz",
       "integrity": "sha512-59vB0RWt09cAct5EIe58+NzGP4TFSD3Bz//2/ELy3ZeTeKF6VTD1AXlH8BGGbCX0PuobZBsIzO7IAI9PH67eKw==",
```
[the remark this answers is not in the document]
awildtechno: package-lock.json is a packages.json file but with the exact order that everything was installed in.

I didn't make any changes to it, but I think the build command may have. It appears to have reverted on its own as well, which is nice.

PR #810, opened 2019-06-03
Add basic 3.7 support

martica:
## PR Checklist:


tests/structures/test_class.py
```diff
@@ -266,3 +266,18 @@ def f(self, a, b, c):
...
+              def __init__(self, value):
+                self.value = list(value)
+
+              def __matmul__(self, other):
+                return A(x * y for x,y in zip(self.value, other.value))
+
+            x = A([1,2,3])
+            y = A([4,5,0])
+            z = x @ y
+            print(z.value)
```
brutusthebee: At column 27: (W291) trailing whitespace

batavia/builtins/dir.js
```diff
@@ -11,7 +12,12 @@ function dir(args, kwargs) {
         throw new exceptions.TypeError.$pyclass('dir() expected exactly 1 argument (' + args.length + ' given)')
     }
     if (args.length === 0) {
-        return "['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'f', 'x']"
+        // Python 3.7 added __annotations__ for use with dataclass
+        if (!version.earlier(3.7)) {
+            return "['__annotations__', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'f', 'x']"
+        } else {
+            return "['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'f', 'x']"
+        }
```
freakboy3742: This may be a subject for a follow on PR - but is this return value correct? dir() returns a list, not a string... this looks like something that will *pass* tests, but only because the output format of the string value is indistinguisable from the rendered version of a list.

batavia/core/version.js
```diff
@@ -6,7 +6,9 @@ var version = {}
 var magic_map = {}
 magic_map[String.fromCharCode(22, 13, 13, 10)] = [[3, 5], [0xF]]
 magic_map[String.fromCharCode(23, 13, 13, 10)] = [[3, 5, 3], [0xF]]
-magic_map[String.fromCharCode(51, 13, 13, 10)] = [[3, 6, 2], [0xF]]
+magic_map[String.fromCharCode(51, 13, 13, 10)] = [[3, 6], [0xF]]
+magic_map[String.fromCharCode(66, 13, 13, 10)] = [[3, 7], [0xF]]
+magic_map[String.fromCharCode(73, 13, 13, 10)] = [[3, 8], [0xF]]
```
freakboy3742: Could this reference the core constants, rather than re-creating the String.fromCharCode calls?

beekeeper.yml
```diff
@@ -4,23 +4,27 @@ pull_request:
...
+        image: pybee/batavia-py35
         name: Smoke build (Python 3.5)
         profile: hi-cpu
     - full-test:
         subtasks:
             - py3.6:
                 name: Python 3.6 tests
-                task: batavia-py36
+                image: pybee/batavia-py36
+                critical: False
```
freakboy3742: From the look of it, the 3.6 tests don't have to be marked as criticial=False - this PR fixes the inconsistencies.

martica: Good call.

batavia/core/version.js
martica: Sweet.

batavia/builtins/dir.js
martica: Ha! It does pass the tests, even though it is wrong in every case. It's a big change so I'll fill an issue for it and we can fix it separately. Funny how it is so easy to follow the format and never think about how it isn't right.

danyeaw:
@martica Thanks @martica for all your teamwork getting this one finished! :tada:

Issue #803, opened 2019-05-30
Correct repr of exceptions.


martica:
## Expected Behavior
print(repr(ValueError('a'))) should match the output in Python.

Before 3.7
ValueError('a',)

3.7+
ValueError('a')

## Current Behavior
ValueError(a)

Commit in cpython where the behaviour changed:
https://github.com/python/cpython/commit/f8a4c03ede6048022f60a58d5a21b278b78a8a16#diff-b49bb0ee9f29e2c8079f1185cf5ed5d1

Additionally: Batavia's exceptions only accept one argument. Python exceptions accept *args.

awildtechno:
Working on it.

PR #811, opened 2019-06-04
BaseException.__repr__ fixes

awildtechno:
Re: Issue #803 

While testing, I discovered that the current implementation of BaseException.__repr__ had a small issue with a comma, but when I tried to fix and test I found out that it wasn't correctly implemented in many ways.

- New behaviors:
-- Exceptions now can accept any argument type, so long as it has a __repr__
-- Exceptions can handle multiple arguments.
-- Strings are handled separately as they need some occasional escaping, which the string repr implementation doesn't handle in the same way for some reason. Not sure if that's an issue with the string repr.
-- Added questionable comma from #803
-- Multiple items print with no comma: "ValueError(1, 2)"
-- Slapped on a case for JavaScript code throwing a Python exception with a repr-less javascript object.
- Test added that tests all of the above.
- My first implementation relied on the *builtins* and *types* helper functions, but I removed them to avoid circular references.

## PR Checklist:

- [ x ] All new features have been tested
- [ x ] All new features have been documented
- [ x ] I have read the **CONTRIBUTING.md** file
- [ x ] I will abide by the code of conduct


batavia/core/exceptions.js
```diff
@@ -432,17 +477,17 @@ exceptions.ValueError = ValueError.prototype.__class__
...
 }
 ZeroDivisionError.prototype = Object.create(Exception.prototype)
 ZeroDivisionError.prototype.__class__ = new Type('ZeroDivisionError', [Exception.prototype.__class__])
 ZeroDivisionError.prototype.__class__.$pyclass = ZeroDivisionError
 
 exceptions.ZeroDivisionError = ZeroDivisionError.prototype.__class__
 
-var JSONDecodeError = function(msg) {
-    Exception.call(this, 'JSONDecodeError', msg)
+var JSONDecodeError = function() {
```
martica: How do you feel about using the ES6 rest syntax here?

```
var JSONDecodeError = function(...args) {
    Exception.call(this, 'JSONDecodeError', args)
}
```

arguments feels magic and having to do Array.from is unfortunate.

batavia/builtins/repr.js
```diff
@@ -7,7 +7,7 @@ function repr(args, kwargs) {
     if (kwargs && Object.keys(kwargs).length > 0) {
         throw new exceptions.TypeError.$pyclass("repr() doesn't accept keyword arguments")
     }
-    if (!args || args.length !== 1) {
+    if (args === undefined || args.length !== 1) {
```
martica: I'm actually not sure why we are testing args at all here. Line 4 already takes the length, making the assumption that it is defined, so by this point, both `!args` and `args === undefined` will be false.

batavia/core/exceptions.js
```diff
@@ -22,16 +22,61 @@ BaseException.prototype.toString = function() {
...
+                    return msg.__repr__()
+                }
+                return msg.toString()
+            }
+        }
+        
+        if (this.args.length) {
+            output += parse(this.args[0])
+            if (this.args.length == 1) {
+                output += ',' // A wild comma shows up in Python 3.5 and 3.6. Removed in 3.7.
```
martica: You can be future-proof here and wrap this in `if (version.earlier(3.7)) {}`

batavia/core/exceptions.js
```diff
@@ -22,16 +22,61 @@ BaseException.prototype.toString = function() {
...
+                if (msg.includes("'")) {
+                    if (msg.includes('"')) { // example: this.msg = '\'"'
+                        msg = msg.replace("'", "\\'")
+                    } else {
+                        wrap = '"'
+                    }
+                }
+                return wrap + msg + wrap
+            } else {
+                // Avoid circular import by calling repr directly.
```
martica: You can require builtins (and types) inside in the function and not get a circular import. It would only be circular if you require it at the top-level.

If you look at the code in the repr builtin, it is checking for a \_\_call__ property on the \_\_repr__ and using that if it exists before calling the flat \_\_repr__ property.

Usually using callables.call_method would be good here, but repr actually skips some of the call_method code where it uses getattr and getattribute. (https://docs.python.org/3/reference/datamodel.html#special-lookup).

batavia/core/exceptions.js
```diff
@@ -22,16 +22,61 @@ BaseException.prototype.toString = function() {
...
         return ''
     }
 }
 
 BaseException.prototype.__repr__ = function() {
-    if (this.msg) {
-        return this.name + '(' + this.msg + ',)'
+    if (this.args != undefined) {
+        let output = ''
+        let count = 0
```
martica: Looks like count isn't used anymore.

tests/structures/test_exception.py
```diff
@@ -45,3 +45,28 @@ def test_raise_inside_loop(self):
                 except:
                     print(x)
         """)
+
+    def test_repr(self):
+        def escape(s):
```
martica: What about just putting all the test input here literally? There isn't much savings having it generated and it has the overhead for the next person to look at it of having to understand the test.

```
        self.assertCodeExecution("""
            print('>>> print(repr(ValueError(\'\')))')
            print(repr(ValueError('')))
            print('>>> print(repr(ValueError(\'"\')))')
            print(repr(ValueError('"')))
            print('>>> print(repr(ValueError("\'")))')
            print(repr(ValueError("'")))
            print('>>> print(repr(ValueError("\\"\'")))')
            print(repr(ValueError("\"'")))
            print('>>> print(repr(ValueError([1, 2, 3])))')
            print(repr(ValueError([1, 2, 3])))
            print('>>> print(repr(ValueError([1], 1, \'1\')))')
            print(repr(ValueError([1], 1, '1')))
        """")
```

martica: Might want to add the empty case as well here.

martica:
Looking good. Nice job filling out the functionality here.

batavia/builtins/repr.js
awildtechno: !args failed a test where args was an empty string. I'll take a look.

martica: Oh. builtins.repr() should always be called with an array for the args param.

batavia/core/exceptions.js
awildtechno: I thought that was the case, but it seemed like the file that defines the exceptions shouldn't share code with files that throw the exceptions. On the very first run, if something crashes, the error handler for it doesn't crash as well and obscure the cause?

martica: I see your point. I can't say for certain that there wouldn't be an issue there. I am still fleshing out my understanding of all the parts of the system here. 

@freakboy3742 Can you give us some advice here? Is it safe to use builtins inside the `__repr__` method of the exceptions? Should we extract the "call a `__repr__` behaviour" from the builtin to somewhere in core without any dependencies?

batavia/builtins/repr.js
awildtechno: Reverted this to the original. Could still be undefined, for example:

var arr = buggyarraybuilder() // returns undefined because javascript
builtins.repr(arr, {})

arguments length still 2, but args is undefined.

awildtechno:
Is the 3.5 test just failing right now or did I break StopIteration and nothing else?

batavia/core/exceptions.js
```diff
@@ -22,16 +22,61 @@ BaseException.prototype.toString = function() {
 }
 
 BaseException.prototype.__str__ = function() {
-    if (this.msg) {
-        return this.msg
+    if (this.args !== undefined) {
```
martica: This should check for length 0. That's why StopIteration.__str__() is different.

batavia/core/exceptions.js
```diff
@@ -22,16 +22,61 @@ BaseException.prototype.toString = function() {
...
+        if (this.args.length == 1) {
+            return this.args[0].toString()
+        }
+        // Multiple args. Format like a tuple.
+        let result = '(' + arg.toString()
+        for (arg in this.args.slice(1)) {
+            result += ', ' + arg.toString()
+        }
+        return result = ')'
```
martica: This should be result + ')'.

That's why `StopIteration.__str__()` is undefined.

tests/structures/test_exception.py
```diff
@@ -45,3 +45,21 @@ def test_raise_inside_loop(self):
                 except:
                     print(x)
         """)
+
+    def test_repr(self):
```
martica: With the unexpected behaviour of `StopIteration.__str__`, maybe we could use a copy of this test with`repr` changed to `str`.

batavia/core/exceptions.js
martica: And I broke this by changing version. You'll need to merge in master and change this to `!version.as_least(3.7)`.

tests/structures/test_exception.py
```diff
@@ -63,3 +63,21 @@ def test_repr(self):
             print('>>> print(repr(ValueError([1], 1, \\'1\\')))')
             print(repr(ValueError([1], 1, '1')))
         """)
+
+    def test_str(self):
+        self.assertCodeExecution("""
+            print('>>> print(str(ValueError()))')
+            print(repr(ValueError('')))
```
martica: Shouldn't this be `print(str(ValueError()))`?

awildtechno:
Looks like I broke the first test somehow. Investigating.

martica:
Probably worth merging master. We messed with the format stuff in the 3.7
branch, although I'm not sure about that message in particular.

batavia/core/exceptions.js
martica: .at_least was added in #802 and .earlier and .later were removed. You'll need to resolve against that before merging.

batavia/builtins/repr.js
martica: I have a vague plan to provide a single validation method to use for all the places where we currently validate "batavia calling conventions". Confirming both that this first argument is not undefined would be a good addition to that in general.

tests/structures/test_exception.py
martica: Sorry. I wasn't clear that there were two changes I was suggesting, `repr` to `str` and `('')` to `()`.

batavia/core/exceptions.js
```diff
@@ -22,16 +22,64 @@ BaseException.prototype.toString = function() {
...
-    if (this.msg) {
-        return this.msg
+    if (this.args.length) {
+        if (this.args.length === 1) {
+            return this.args[0].toString()
+        }
+        // Multiple args. Format like a tuple.
+        let result = '(' + this.args[0].toString()
+        for (let arg in this.args.slice(1)) {
+            result += ', ' + arg.toString()
```
martica: That expected failure on the new test was niggling at me so I took at look. It's the very unfortunate JS `for...in` vs `for...of` that is primary cause, but additionally, it looks like BaseException's `__str__` calls `__repr__` on the args when there are multiple instead of `__str__`. You could probably move the parse function you defined in `__repr__` and reuse that.

```
>>> class Blah:
...   def __str__(self):
...     return "StrBlah"
...   def __repr__(self):
...     return "ReprBlah"
...
>>> str(ValueError(Blah(), Blah()))
'(ReprBlah, ReprBlah)'
>>> str(ValueError(Blah()))
'StrBlah'
```

martica:
Everything is looking good. I'll make the sync against master and merge it.

batavia/core/exceptions.js
martica: Very nice clean up for this, using the tuple code.

batavia/core/exceptions.js
martica: It's okay to leave this as it is, so I'll resolve this.

batavia/core/exceptions.js
martica: I've done this change myself now.

martica:
Thanks for all the work on this. We are in a much better state for exceptions now.

PR #817, opened 2019-06-13
Add list tests

alexjdw:
Noticed an incorrect error message in builtins/list, then noticed that test_list had no tests, so I made some. Working on getting them passing in a separate commit.

Let me know if you can think of any more cases.


martica:
It looks like a lot of these tests are duplicated in tests/datatypes/test_list.py to some extent, at least the parts that are passing. How do you feel about moving them into that file? At least the ones that don't directly concern the list() builtin function.

phildini:

PR #819, opened 2019-06-13
How-To Update for docs.

awildtechno:
Docs felt a bit rusty. I decided to add some pages. I tried to think of what I would've liked to know on my first commit, and I included some of the questions I had to have addressed by the team. Please let me know what you think and if there's any sections that could be added/improved.

- "tour of the code" added with basic project structure and a list of helpful functions
- builtins guide improved and geared toward new contributors
- added a similar guide for types geared toward new contributors
- broke out test contributions into its own guide
- expanded on it with more examples and an overview of the cleaners.
- How-to index updated to include the new stuff.

Possible issues:

- A bit repetitive between the three updated files.
- Very wordy, so tl;dr risk?

### DIFF
--- a/docs/how-to/builtins.rst
+++ b/docs/how-to/builtins.rst
@@ -1,9 +1,9 @@
 Implementing Python Built-ins in JavaScript
 ===========================================
 
-Pythons builtins are the absolute core of the language. If you're new to Python, read up on them here: https://docs.python.org/3/library/functions.html
+Pythons builtins give Python its trademark magical feel. If you're new to Python, read up on them here: https://docs.python.org/3/library/functions.html
 
-Most builtins have already been added to the project, but many do not follow the correct spec outlined below.
+Most builtins have already been added to the project, but many do not follow the spec outlined below.
 Some may not handle certain types of inputs correctly. In addition, new builtins may of course arrive with the
 latest and greatest Python version. This guide should serve as your field manual for adding, updating, and
 navigating our implementations.
--- a/docs/how-to/builtins.rst
+++ b/docs/how-to/builtins.rst
@@ -50,7 +50,7 @@ JavaScript object. The first thing to do is check that both were passed in.
         }
 
 Next, we need to validate the arguments are correct. We can use JavaScript's ``Object.keys()`` to
-get the keys of an object. If the function pass in the , we will check the Python REPL to see
+get the keys of an object. If we can't accept certain args or kwargs, we will check the Python REPL to see
 what kind of error and throw it.
 
 .. tabs::
--- a/docs/how-to/builtins.rst
+++ b/docs/how-to/builtins.rst
@@ -4,7 +4,7 @@ Implementing Python Built-ins in JavaScript
 Pythons builtins give Python its trademark magical feel. If you're new to Python, read up on them here: https://docs.python.org/3/library/functions.html
 
 Most builtins have already been added to the project, but many do not follow the spec outlined below.
-Some may not handle certain types of inputs correctly. In addition, new builtins may of course arrive with the
+Some may not handle certain types of inputs correctly. In addition, new builtins may arrive with the
 latest and greatest Python version. This guide should serve as your field manual for adding, updating, and
 navigating our implementations.
 
--- a/docs/how-to/builtins.rst
+++ b/docs/how-to/builtins.rst
@@ -1,107 +1,125 @@
 Implementing Python Built-ins in JavaScript
 ===========================================
 
+Process
+-------
+
+The first thing to do when adding anything to Batavia is to play around a bit with it in the Python REPL::
+
+    >> list()
+    []
+    >> []
+    []
+    >> list(4)
+    Traceback (most recent call last):
+    File "<stdin>", line 1, in <module>
+    TypeError: 'int' object is not iterable
+
+Your goal is to find out how the function responds to various inputs and outputs. You may also
+want to consult the offical documentation. Once you're a little familiar, you can add your
+implementation to Batavia.
+
 General Structure
------------------
+*****************
 
 JavaScript versions of Python built-in functions can be found inside the ``batavia/builtins``
-directory in the Batavia code. Each built-in is placed inside its own file.
+directory in the Batavia code. Each built-in is placed inside its own file. These builtins are 
+designed to be used only inside Batavia, as such they need to ensure
+they are being used in a compatible manner.
+
+Each builtin function will receive arguments and keyword arguments and needs to handle them,
+even if the result is throwing an error. Args should be an array, and kwargs should be a 
+JavaScript object. The first thing to do is check that both were passed in.
 
 .. code-block:: javascript
 
     // Example: a function that accepts exactly one argument, and no keyword arguments
 
-    var <fn> = function(<args>, <kwargs>) {
-        // These builtins are designed to be used only inside Batavia, as such they need to ensure
-        // they are being used in a compatible manner.
-
-        // Batavia will only ever pass two arguments, args and kwargs. If more or fewer arguments
-        // are passed, then Batavia is being used in an incompatible way.
-        // See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments
+    var <fn> = function(args, kwargs) {
+        // Always add this code.
         if (arguments.length !== 2) {
             throw new builtins.BataviaError.$pyclass("Batavia calling convention not used.");
         }
 
-        // We are now checking if a kwargs object is passed. If it isn't, kwargs will be null. Like
-        // obj.keys() in Python we can use Object.keys(obj) to get the keys of an object. If the
-        // function doesn't need support any kwargs we throw an error.
-        if (kwargs && Object.keys(kwargs).length > 0) {
-            throw new builtins.TypeError.$pyclass("<fn>() doesn't accept keyword arguments.");
-        }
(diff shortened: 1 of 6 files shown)


docs/how-to/builtins.rst
```diff
@@ -1,104 +1,138 @@
 Implementing Python Built-ins in JavaScript
 ===========================================
 
+Python's builtins give Python its trademark magical feel. If you're new to Python, read up on them here: https://docs.python.org/3/library/functions.html
```
freakboy3742: This would be better as an inline link, rather than a raw URL

docs/how-to/builtins.rst
```diff
@@ -1,104 +1,138 @@
...
+
+Most builtins have already been added to the project, but many do not follow the spec outlined below.
```
freakboy3742: Avoid colloquial abbreviations; specification, not spev. Although in this case, "description" is probably more accurate; it's not really a formal spec.

docs/how-to/builtins.rst
```diff
@@ -1,104 +1,138 @@
...
+even if the result is throwing an error. Args should be an array, and kwargs should be a 
+JavaScript object. The first thing to do is check that both were passed in.
 
-    // Example: a function that accepts exactly one argument, and no keyword arguments
+.. code-block:: javascript
 
-    var <fn> = function(<args>, <kwargs>) {
-        // These builtins are designed to be used only inside Batavia, as such they need to ensure
-        // they are being used in a compatible manner.
+    // Example: <fn> accepts exactly one argument, and no keyword arguments
```
freakboy3742: Since we're cleaning this up... I'm not wild about using "<fn>" as an example. It's potentially confusing, since as written, it's not valid Javascript *or* Python syntax. I'd rather use an actual concrete example, with a preface note that says "we're going to use abs() as an example here; abs() is already implemented, but the pattern will be the same for any other builtin". (Substitute abs() for whatever builtin makes sense in the content)

docs/how-to/builtins.rst
```diff
@@ -1,104 +1,138 @@
...
+* If you are throwing an error (excluding ``BataviaError``) anywhere, write a test that tries to throw it.
+* If you accounted for an edge case (look for an ``if`` statement), test it.
+* Check out the official documentation (https://docs.python.org/3/) for more edge cases.
 
-Take a guess at the implementation structure based on the other functions.
+Credit
+^^^^^^
 
-Copy the style of the other implemented functions
+Thanks to Kshitij Sobti for the original document that this guide is based on. Most code examples are from his original.
```
freakboy3742: I appreciate the sentiment here of recognizing past authors, but it's a little out of place in main project documentation.

docs/how-to/contribute-tests.rst
```diff
@@ -0,0 +1,118 @@
...
+
+                def __bool__(self):
+                    return self.val == 1
+            print(bool(BoolLike(0)))
+            print(bool(BoolLike(1)))
+            """)
+
+The ``assertCodeExecution`` method will run the code provided to it in both implementations. This
+code needs to generate some output so that the output can be compared, hence the need to print the
+values. **Code that is not being printed is not being tested.**
```
freakboy3742: Do we want to mention some of the edge cases, like code that produced equivalent output in a different order (e.g., sets and dict keys) will also fail, so tests should be structured to compensate - either by restructuring the test (e.g., convert set to a sorted list and compare the type of the original value, rather than using set comparison), or using test output subsitutions?

docs/how-to/contribute-tests.rst
```diff
@@ -0,0 +1,118 @@
...
+        code = """
+            try:
+                code_that_raises_a_ValueError()
+            except ValueError as err:
+                print(err)
+        """
+        self.assertCodeExecution(code)
+
+Remember to catch the specific error you want, and then print the error. **Code that is not being printed
+is not being tested.**
```
freakboy3742: Also: We need to output something on *success* as well - because we need to differente between "succeeded" and "failed but then failed again when printing the error".

docs/how-to/index.rst
```diff
@@ -12,5 +12,8 @@ How-to guides are recipes that take the user through steps in key subjects. They
 
    Contribute to Batavia's code <contribute-code>
    Contribute to Batavia's documentation <contribute-docs>
+   Tour of the Code <tour>
+   Follow the Goat: Batavia's Test Suite <contribute-tests>
```
freakboy3742: This is a cute in-joke (I lol'd :-) but if you don't know about the testing goat, this is going to be a weird title. As much as I love puns, in this case, I think it detracts from the purpose. (plus, it's *obey* the testing goat... :-P )

docs/how-to/tour.rst
```diff
@@ -0,0 +1,122 @@
...
+important features of the code. It's aimed at new contributors, but frequent flyers can
+skip down to the bottom for a quick reference.
+
+General Structure
+-----------------
+
+Core Code
+*********
+
+Batavia implements the Python native types, builtins, and standard library partially in JavaScript.
```
freakboy3742: This could be a a little clearer - "the core language (types, builtins and the core interpreter) are all javascript, as well those portions of the standard library that depend on system services".

docs/how-to/tour.rst
```diff
@@ -0,0 +1,122 @@
...
+This ensures quick execution and less compiling of the compiler. These are orgainzed into the
+``builtins``, ``core``, ``modules``, ``stdlib``, and ``types`` files of the main ``/batavia`` directory, with
+corresponding subdirectories. Alongside the virtual machine and test suite, this code makes 
+up the bulk of Batavia. New contributors should start with the ``types`` and ``builtins`` sections
+for the best examples to review and copy from. These implementations are the foundation of Python as you know it and
+should be immediately familiar to a Python developer.
+
+Support
+*******
+You'll also notice folders for tests, docs, and a few other bits and bobs, like ``testserver``, which is
```
freakboy3742: Avoid the colloquial - "bits and bobs"

docs/how-to/tour.rst
```diff
@@ -0,0 +1,122 @@
...
+a Django project that allows code execution from your browser. Contributions to the
```
freakboy3742: might be better to describe testserver as a sample deployment that lets you test Batavia.

docs/how-to/tour.rst
```diff
@@ -0,0 +1,122 @@
...
+
+.. code-block:: javascript
+
+    var builtins = require('./builtins.js')
+    builtins.abs([-1], {})              // equivalent to abs(-1)
+
+This contains all of the native Python builtin functions, like ``str``, ``len``, and ``iter``.
+
+When dealing with Python types, many of the native JavaScript operations have been modified to
+try to use builtins first. For instance, .toString() will often just call the object's __str__ if
```
freakboy3742: code markup on tostring and __str__

docs/how-to/tour.rst
```diff
@@ -0,0 +1,122 @@
...
+
+This allows us to avoid ugly things like comparing ``Object.prototype.constructor``. Instead,
+use ``types.isinstance`` or ``types.isbataviainstance``. Secondly, it's important that the inputs to Python
+types are Pythonized themselves where needed. You should not be making a list() of JavaScript arrays, for
+instance. That doesn't make sense! (It may even pass some tests, which is dangerous.)
+
+core/callables.js
+^^^^^^^^^^^^^^^^^
+
+These methods ensure that all Python code is executed using the proper __call__ procedure, which could be
```
freakboy3742: markup on __call__

docs/how-to/tour.rst
```diff
@@ -0,0 +1,122 @@
...
+of ``call_function`` instead of calling Python functions and methods directly. An example:
+
+.. code-block:: javascript
+
+    // Avoid this
+    my_thing.__repr__()
+
+    // Better
+    var callables = require('./core/callables.js')
+    callables.call_method(my_thing, '__repr__', [], {})
```
martica: This is actually a bit of a special case. `__repr__` is one of the methods that isn't looked up with getattribute() in cpython. I haven't done a lot of spelunking here yet, and the docs are vague on which methods are in this list, but `__repr__` is explicitly listed.

https://docs.python.org/3/reference/datamodel.html#special-method-lookup

docs/how-to/contribute-tests.rst
awildtechno: Added a bit for this.

docs/how-to/types.rst
```diff
@@ -0,0 +1,152 @@
...
+your pull request on GitHub to run the CI. Fix any issues and push again until CI passes. The Batavia team will get back
+to you with any additional notes and edits.
+
+Tips
+^^^^
+
+Your goal is to mimic the CPython implementation as much as possible. If you do so, you'll often fix multiple issues at once. Here's some tips:
+
+* The original implementation is documented in detail at https://docs.python.org/3/ -- reading up there will definitely improve your understanding.
+* If you're inherting your class from JavaScript, which is very common, you get JavaScript's internal methods for free. Oftentimes, they can be left as is or lightly wrapped.
```
pzrq: @awildtechno Tiny nitpick: `inherting` I think could be `inheriting`?

I'm also not sure if it makes sense to add something referring to tox / tox.ini or https://github.com/beeware/beekeeper (could be in a later PR though)?

Otherwise in my quick overview before sleep, this PR looks awesome and way better than what I just started writing, will definitely enjoy going through it in more depth when PyConAU dev sprints continue tomorrow!

Issue #828, opened 2019-10-05
The links to contributor's guide and others appear to be broken


pranavmodx:
3 of the links given in the the section : 'docs/how-to/contribute-code.rst' appear to be broken. I could help fix this by replacing them with the correct links if someone could point me to where they are obtained from.

freakboy3742:
Thanks for the report! These links obviously got broken during a recent docs refactor.

The correct destinations should be:
* "Separate contributors guide" -> The contribute-docs.rst file in the same folder
* "Getting started guide" -> tutorial 0
* "Set up your development environment" -> the development-env.rst file in the same folder

For bonus points; adding a build test that validates that documentation is accurate would be extremely helpful.

pranavmodx:
I have updated the links to correct destinations. 
I looked at the test sections but can't seem to figure out a way for adding a build test or how it works. Could you please guide me through it?

freakboy3742:
So - Adding a build test would mean adding a CI stage that does a build of the documentation to confirm that there are no errors. We're in the process of moving all our CI to Github Actions; if you wanted to try adding a Github actions configuration for Batavia, replacing the current BeeKeeper configuration, but adding a documentation step, that would be really helpful. 

As a point of reference, [Briefcase](github.com/beeware/briefcase) has a [Github Actions configuration in it's v0.3 branch]([https://github.com/beeware/briefcase/blob/v0.3/.github/workflows/ci.yml])

PR #829, opened 2019-10-08
Fixed some broken links in 'docs/how-to' section

pranavmodx:
- Updated the links given in the documentation to the correct destinations.
- The links are directed to the desired pages and 'page not found' errors are eliminated.
- Refs #828 

### DIFF
--- a/docs/how-to/contribute-code.rst
+++ b/docs/how-to/contribute-code.rst
@@ -9,13 +9,13 @@ what you don’t know!
 
 This tutorial is also going to focus on code contributions. If your interests
 and skills are in documentation, `we have a separate contributors guide
-<contribute-docs>`__ just for you.
+<contribute-docs.rst>`__ just for you.
 
 Do the tutorial first!
 ----------------------
 
 Before you make your first contribution, take Batavia for a spin. The
-instructions in the `getting started guide </tutorial/tutorial-0>`__ *should*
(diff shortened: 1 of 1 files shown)


freakboy3742 (changes requested):
I'm not sure how you tested these, but I'm not seeing any change when I try these modifications locally. Sphinx doesn't require referencing the `.rst` extension in it's links...

pranavmodx:
I tested them from the preview feature of GitHub itself. It seems to be working fine.

What I've observed is that the links before my changes were of the format - <main_url>/docs/how-to/<file_name_without_extension> and these didn't work

After inserting the extension in the links to get the format - <main_url>/docs/how-to/<file_name_with_extension> , they worked perfectly fine. 
When the reference '.rst' is added in the links, they redirect to links of this working format.

Also, only 1 link has the wrong path (Tutorial 0). The other two have the right paths but the extension referencing is not there.

freakboy3742:
Github doesn't do a very good job of rendering ReStructured Text, and doesn't render Sphinx directives at all. Github's rendering isn't a useful reference in this case.

To verify the rendering, you need to use Sphinx to build the documentation, like it says in the [docs contribution guide](https://batavia.readthedocs.io/en/latest/how-to/contribute-docs.html).

phildini:

